Issue #18, opened 2025-08-05
Tests seem to be still flaky


sebst:
See: https://github.com/michal-stlv/stelvio/actions/runs/16744271209/job/47398988280?pr=17

michal-stlv:
hmmmm. i can't make them fail on main. looking out that job, it fails on asserting role which is basic stuff always created during api gateway deployment. btw. this is the tests added where I fixed #9 . can you make it fail locally on your branch? is it like 1 of 5 or 1 of 10 times  or  even less often?

michal-stlv:
I was able to reproduce this locally together with deployment issues where some resources are not deployed in the right order causing deployment to fail. I Identified few places which are causing this and am working on the fix.

michal-stlv:
Closing this as I fixed it in #26 . If it appears again we can reopen.

PR #21, opened 2025-08-08
WIP: Try to isolate components between tests

sebst:
This is WIP!

### DIFF
--- a/tests/aws/pulumi_mocks.py
+++ b/tests/aws/pulumi_mocks.py
@@ -62,7 +62,8 @@ class PulumiTestMocks(Mocks):
         elif args.typ == "aws:apigateway/resource:Resource":
             output_props["id"] = f"resource-{args.name}"
         elif args.typ == "aws:apigateway/account:Account":
-            ...
+            # Mock API Gateway account with no cloudwatch_role_arn initially
+            output_props["cloudwatch_role_arn"] = args.inputs.get("cloudwatch_role_arn", "")
         elif args.typ == "aws:dynamodb/table:Table":
             output_props["arn"] = f"arn:aws:dynamodb:{region}:{account_id}:table/{name}"
         # LayerVersion resource
--- a/tests/aws/acm/test_acm_validated_domain.py
+++ b/tests/aws/acm/test_acm_validated_domain.py
@@ -134,9 +134,11 @@ def pulumi_mocks():
 def component_registry():
     ComponentRegistry._instances.clear()
     ComponentRegistry._registered_names.clear()
+    ComponentRegistry._user_link_creators.clear()
     yield ComponentRegistry
     ComponentRegistry._instances.clear()
     ComponentRegistry._registered_names.clear()
+    ComponentRegistry._user_link_creators.clear()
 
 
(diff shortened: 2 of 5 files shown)


sebst:
> The test_empty_api test was failing because the API Gateway account mock was incomplete. The _create_api_gateway_account_and_role() function uses Account.get() to reference an existing API Gateway account, but the Pulumi mock for aws:apigateway/account:Account was incomplete (just had ...).

> The issue is likely a race condition or inconsistent test execution order where:

> Some tests run with the context from [conftest.py](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (no DNS provider)
> Some tests run with contexts set up by other test fixtures that include DNS providers
> The context state might not be properly isolated between tests
> The flakiness suggests that test isolation is broken - tests are interfering with each other's context state.

tests/aws/pulumi_mocks.py
```diff
@@ -62,7 +62,8 @@ def new_resource(self, args: MockResourceArgs) -> tuple[str, dict[str, Any]]:  #
         elif args.typ == "aws:apigateway/resource:Resource":
             output_props["id"] = f"resource-{args.name}"
         elif args.typ == "aws:apigateway/account:Account":
-            ...
+            # Mock API Gateway account with no cloudwatch_role_arn initially
+            output_props["cloudwatch_role_arn"] = args.inputs.get("cloudwatch_role_arn", "")
```
michal-stlv, line 66: if no cloudwatch_role_arn in inputs we should assign None, not "", or not assign at all.

michal-stlv:
> > The test_empty_api test was failing because the API Gateway account mock was incomplete. The _create_api_gateway_account_and_role() function uses Account.get() to reference an existing API Gateway account, but the Pulumi mock for aws:apigateway/account:Account was incomplete (just had ...).
> 
> > The issue is likely a race condition or inconsistent test execution order where:
> 
> > Some tests run with the context from conftest.py (no DNS provider)
> > Some tests run with contexts set up by other test fixtures that include DNS providers
> > The context state might not be properly isolated between tests
> > The flakiness suggests that test isolation is broken - tests are interfering with each other's context state.

@sebst yes correct but this unfortunately only part of problem, and I suspect it will not solve the whole thing. There are some dependencies issues when creating gateway - also with method and integrations so order of operation is currently not guaranteed hence this error. I also got these error during deployment.  I mentioned it in #18

michal-stlv:
@sebst I fixed flaky tests in PR  #26 but here you also do some other stuff in tests, namely cleaning/resseting some global values. if you think that's needed you can merge dev back to your branch and we can merge it. if not let's close this?

sebst:
> @sebst I fixed flaky tests in PR #26 but here you also do some other stuff in tests, namely cleaning/resseting some global values. if you think that's needed you can merge dev back to your branch and we can merge it. if not let's close this?

I'm fine with #26 .

Let's keep this one open for a couple of days to see whether problems re-surface, and this might help

PR #25, opened 2025-08-08
PoC: Basic S3 support

### LINKED ISSUE #23: S3 Support
As a user, I want to create S3 buckets with Stelvio.
On creation, the following attributes should be given:
- Region
- Visibility
- Storage Class

As a developer, I want to access S3 buckets from within my serverless function
- Using Stelvio's link functionality, I need access to credentials and URLs of the S3 bucket
- I can use any (preferably `boto3` library within my functions to interact with the bucket

sebst:
DRAFT!!!!

Closes #23


stelvio/aws/s3.py
```diff
@@ -0,0 +1,61 @@
...
+    @property
+    def arn(self) -> Output[str]:
+        return self.resources.bucket.arn
+
+    def _create_resources(self) -> S3BucketResources:
+        bucket = pulumi_aws.s3.Bucket(
+            context().prefix(self.name),
+        )
+        pulumi.export(f"s3bucket_{self.name}_arn", bucket.arn)
+        pulumi.export(f"s3bucket_{self.name}_name", bucket._name)  # noqa: SLF001
```
michalmartinka, line 37: do we need to access `_name` ?  what is in there? don't we have the value in self.name or is it same as `context().prefix(self.name)` or is it name generated by pulumi so it will have -123456 suffix? In that case we should be `bucket.name` without underscore to get actually bucket name created on aws.

michal-stlv: do we need to access _name ? what is in there? don't we have the value in self.name or is it same as context().prefix(self.name) or is it name generated by pulumi so it will have -123456 suffix? In that case we should be bucket.name without underscore to get actually bucket name created on aws.

sebst:
This PR can be closed in favour of #31. Keep the branch for future work (as the static website and cloudfront stuff should be handled separately.

PR #30, opened 2025-08-14
feat: truncate role and policy name to fit aws limits

### LINKED ISSUE #29: Hash part of the resource name if it's longer than allowed by cloud provider
When Stelvio creates names for generated resources, namely role and policy these name might end up being longer than allowed by cloud provider (64 and 128) Stelvio should make sure they fit the limit to make sure deployment is possible.

Reported also in #13

michal-stlv:
Closes #29

### DIFF
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -326,13 +326,13 @@ Diff for stelvio-app → michal
 
 + to create  stelvio-app-michal-todos → aws:dynamodb/table:Table
 + to create  stelvio-app-michal-todos-api → aws:apigateway/restApi:RestApi
-+ to create  stelvio-app-michal-todos-api-functions-todos-policy → aws:iam/policy:Policy
-+ to create  stelvio-app-michal-todos-api-functions-todos-role → aws:iam/role:Role
++ to create  stelvio-app-michal-todos-api-functions-todos-p → aws:iam/policy:Policy
++ to create  stelvio-app-michal-todos-api-functions-todos-r → aws:iam/role:Role
 + to create  stelvio-app-michal-resource-todos → aws:apigateway/resource:Resource
-+ to create  stelvio-app-michal-todos-api-functions-todos-basic-execution-role-policy-attachment → aws:iam/rolePolicyAttachment:RolePolicyAttachment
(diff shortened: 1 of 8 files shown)


sebst:
In general, shouldn't `safe_name` always be used? Do you think we could integrate it in context? Something like `context().gen_name(name, basic-execution-r-p-attachment)` instead of `safe_name(context().prefix(), name, 64, "-r")`?

Also, could we come up with more meaningful short names?
E.g.
`default-role-policy-attachment` --> `def-role-policy-attach` instead of `default-r-p-attachment`?

We should document the naming styles

michal-stlv:
> In general, shouldn't `safe_name` always be used? 

Well this is complicated. Generally that's what I wanted to do but it's hard to find what are name limits for different aws resoruces. It's 64 for role 128 for policy. But then e.g role policy attachment is not aws resource, only pulumi concept.  Then others we generate e.g. api gateway resource, method and integrations seems to have generous limits so should be ok.

And that's currently all resources that Stelvio creates automatically. For user named resources I don't want to truncate them, it's up to user to change them if too long I'd say.

So I'm happy to do more on this when we find it's actually an issue. 

> Do you think we could integrate it in context? Something like `context().gen_name(name, basic-execution-r-p-attachment)` instead of `safe_name(context().prefix(), name, 64, "-r")`?

I initially had safe_name on context... yeah maybe we can have `context().prefix_safe()` or `context().gen_name()` or something. 


> Also, could we come up with more meaningful short names? E.g. `default-role-policy-attachment` --> `def-role-policy-attach` instead of `default-r-p-attachment`?
 
`-r-p-attachment` is shortcut for role policy attachment. i want to be consistent with `-r` and `-p` shortcuts for role and policy 

> We should document the naming styles

Yes, will do.

sebst:
> And that's currently all resources that Stelvio creates automatically. For user named resources I don't want to truncate them, it's up to user to change them if too long I'd say.

The user-defined names get modified (prefixed, at least), right?
So, if there's a modification, we could (and should?) also truncate it. 
Somewhat strange to say "don't use names longer than 64-12 (or whatever the prefix length is) characters", no?

To me, the PR is ok for now, but we should have a centralised name generator/truncator/validator in the future. Let's create a new ticket for that.

Just as an idea, we could also have a mapping of Stelvio names (no limit) to Pulumi names (truncated). But again, that's for the future.

PR #31, opened 2025-08-22
Initial base setup for buckets based on #25.

### LINKED ISSUE #23: S3 Support

sebst:
Closes #23

### DIFF
--- a/stelvio/aws/s3.py
+++ b/stelvio/aws/s3.py
@@ -1,0 +1,83 @@
+from dataclasses import dataclass
+from typing import final
+
+import pulumi
+import pulumi_aws
+
+from stelvio import context
+from stelvio.aws.permission import AwsPermission
+from stelvio.component import Component, ComponentRegistry, link_config_creator
+from stelvio.link import Link, Linkable, LinkConfig
+
+
+@dataclass(frozen=True)
+class S3BucketResources:
+    bucket: pulumi_aws.s3.Bucket
+
+
+@final
+class Bucket(Component[S3BucketResources], Linkable):
+    def __init__(
+        self,
+        name: str,
+        storage_class: str = "STANDARD",
+        versioning_enabled: bool = False,
+        public_access_block: bool = True,
+        enforce_https: bool = True,
+    ):
+        super().__init__(name)
+        self.storage_class = storage_class
+        self.versioning_enabled = versioning_enabled
+        self.public_access_block = public_access_block
+        self.enforce_https = enforce_https
+        self._resources = None
+
+    def _create_resources(self) -> S3BucketResources:
+        bucket = pulumi_aws.s3.Bucket(
+            context().prefix(self.name),
+            versioning={"enabled": self.versioning_enabled},
+        )
+
+        # Configure public access block
+        public_access_block = pulumi_aws.s3.BucketPublicAccessBlock(
+            context().prefix(f"{self.name}-pab"),
+            bucket=bucket.id,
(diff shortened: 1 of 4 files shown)


stelvio/aws/s3.py
```diff
@@ -0,0 +1,83 @@
...
+        self,
+        name: str,
+        storage_class: str = "STANDARD",
+        versioning_enabled: bool = False,
+        public_access_block: bool = True,
+        enforce_https: bool = True,
+    ):
+        super().__init__(name)
+        self.storage_class = storage_class
+        self.versioning_enabled = versioning_enabled
```
michal-stlv, line 42: As discussed we can do `Literal["STANDARD", "OTHER_VALUE"]`

stelvio/aws/s3.py
```diff
@@ -0,0 +1,83 @@
...
+        self.public_access_block = public_access_block
```
michal-stlv: What happens when public is set to False? We need to explain user in the docs and also maybe we need some middle ground - other settings to saet what is public/private etc. bucket/objects etc.

stelvio/aws/s3.py
sebst: Left to the user -- updated documentation

stelvio/aws/s3.py
sebst: Now completely customizable, updated doc

stelvio/aws/s3.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@final
+class Bucket(Component[S3BucketResources], Linkable):
+    def __init__(
+        self,
+        name: str,
+        versioning_enabled: bool = False,
+        prevent_public_access: bool = True,
+        public_access_block_args: dict | None = None,
```
sebst: Todo: better typing

stelvio/aws/s3.py
```diff
@@ -0,0 +1,83 @@
...
+        self.versioning_enabled = versioning_enabled
+        self.public_access_block = public_access_block
+        self.enforce_https = enforce_https
+        self._resources = None
+
+    def _create_resources(self) -> S3BucketResources:
+        bucket = pulumi_aws.s3.Bucket(
+            context().prefix(self.name),
+            versioning={"enabled": self.versioning_enabled},
+        )
```
michal-stlv: What will be name of the actual bucket? According to docs if no parameter `bucket` is set then provider will generate random name: https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucket/#inputs

stelvio/aws/s3.py
```diff
@@ -0,0 +1,83 @@
...
+
+@final
+class Bucket(Component[S3BucketResources], Linkable):
+    def __init__(
+        self,
+        name: str,
+        storage_class: str = "STANDARD",
+        versioning_enabled: bool = False,
+        public_access_block: bool = True,
+        enforce_https: bool = True,
```
michal-stlv: I think if we are over 3-4 params here (except name) we should use Config object as we do for `Function` <> `FunctionConfig`. I'm now doing same for DynamoTable. So in this case we'd have `BucketConfig` which would have those properties.

PR #33, opened 2025-08-23
feat: s3 buckets support

sebst:


stelvio/aws/s3.py
```diff
@@ -0,0 +1,80 @@
...
+class Bucket(Component[S3BucketResources], Linkable):
+    def __init__(
+        self,
+        name: str,
+        versioning_enabled: bool = False,
+        prevent_public_access: bool = True,
+    ):
+        super().__init__(name)
+        self.versioning_enabled = versioning_enabled
+        self.prevent_public_access = prevent_public_access
```
michal-stlv: Can we make the name of the param shorter? Just `access` with default `None` (meaning no access at all -completely private) and if user wants to turn it of he'd set it to `"private"` then it would be turned into readonly - meaning it would be read only access to the all files in the bucket. This would also need to be documented and explained in the docs. (We need to research exact configuration for that to make sure that it is properly and tightly configured according to best practices ). 
Param value would be then `access: Literal["public"] | None = None`

stelvio/aws/s3.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # Configure public access block
+        public_access_block = pulumi_aws.s3.BucketPublicAccessBlock(
+            context().prefix(f"{self.name}-pab"),
+            bucket=bucket.id,
+            block_public_acls=self.prevent_public_access,
+            block_public_policy=self.prevent_public_access,
+            ignore_public_acls=self.prevent_public_access,
+            restrict_public_buckets=self.prevent_public_access,
+        )
```
michal-stlv: Here I'd just go with 
```py
if self. access == "public":
    # setup readonly configuration 
else:
    pulumi_aws.s3.BucketPublicAccessBlock(
            context().prefix(f"{self.name}-pab"),
            bucket=bucket.id,
            block_public_acls=True,
            block_public_policy=True,
            ignore_public_acls=True,
            restrict_public_buckets=True,
        )
```

michal-stlv: looks like for readonly we need to have `BucketPublicAccessBlock` values set all to `False` and also create bucket policy:
```py
public_read_policy = get_policy_document_output(
        statements=[{
            "effect": "Allow",
            "principals": [{
                "type": "*",
                "identifiers": ["*"],
            }],
            "actions": ["s3:GetObject"],
            "resources": [self.bucket.arn.apply(lambda arn: f"{arn}/*")],
        }]
    )
    
pulumi_aws.s3.BucketPolicy(
    context().prefix(f"{self.name}-policy"),
    bucket=self.bucket.id,
    policy=public_read_policy.json
)
```

But I haven't done much reaserach needs to be checked and documented.

tests/aws/s3/test_s3_bucket.py
```diff
@@ -0,0 +1,189 @@
...
+    _ = bucket.resources
+
+    # Assert
+    def check_resources(args):
+        bucket_id, arn = args
+        assert bucket_id == TP + "my-bucket-test-id"
+        assert arn == BUCKET_ARN_TEMPLATE.format(name=tn(TP + "my-bucket"))
+
+    pulumi.Output.all(bucket.resources.bucket.id, bucket.arn).apply(check_resources)
+
```
michal-stlv, line 37: We need tests that  also validates the following:
- no versioning is enabled if `versioning_enabled` is false
- versioning is enable  if `versioning_enabled` is True
- if access is None then tests that PAB is created with True flags and policy is not created
- if access is `public` test that PAB is craeted with False flags and policy is created with proper values

stelvio/aws/s3.py
```diff
@@ -0,0 +1,112 @@
...
+class S3BucketResources:
+    bucket: pulumi_aws.s3.Bucket
+    public_access_block: pulumi_aws.s3.BucketPublicAccessBlock
+    bucket_policy: pulumi_aws.s3.BucketPolicy | None
+
+
+@final
+class Bucket(Component[S3BucketResources], Linkable):
+    def __init__(
+        self, name: str, versioning_enabled: bool = False, access: Literal["public"] | None = None
```
michal-stlv: `versioning_enabled` is quite long .Could it be maybe just `versioning`?

michal-stlv (changes requested):
Very nice. Just needs few more tests and consider that name change.

tests/aws/s3/test_s3_bucket.py
```diff
@@ -0,0 +1,333 @@
...
+        buckets = pulumi_mocks.created_s3_buckets(TP + "test-bucket")
+        assert len(buckets) == 1
+        created_bucket = buckets[0]
+
+        # Check that versioning is explicitly set to enabled
+        versioning_config = created_bucket.inputs.get("versioning")
+        assert versioning_config is not None
+        assert versioning_config["enabled"] is True
+
+    bucket.resources.bucket.id.apply(check_versioning)
```
michal-stlv: These two tests are good but basically same so coudl be parametrized ?

PR #37, opened 2025-09-08
docs: small docs updates dns, bucket, dynamo

michal-stlv:


docs/guides/dns.md
```diff
@@ -56,7 +54,11 @@ This example initializes a Stelvio application with Cloudflare DNS. You need to
     )
     ```
 
-    This class will handle the creation and validation of the TLS certificate for your custom domain. You can then use this domain in your Stelvio application.
+    This class will handle the creation and validation of the TLS certificate for your custom domain.  
+    You can then use this domain in your Stelvio application.
+    
+    **However, Stelvio usually handles this step for you (e.g. when using custom domain with API Gateway)**
```
michal-stlv, line 60: Hey @sebst I added this as I don't think users need to use `AcmValidatedDomain` themselves ? It's all handled for them no? So technically right now it's just internal thing? Unless they'd want to create validated domain and withotu using it in Stelvio?

PR #40, opened 2025-09-12
feat: dynamodb table streams

### LINKED ISSUE #38: Support DynamoDB streams
DynamoDb Streams allows you to react to changes of the data stored in the table. So every time item in the table is created, updated or deleted event is triggered and you can subscribe to it and perform some action. E.g. every time user is created your lambda can be called and you can send email. 

Stelvio wants to support this setup and make it very easy. 

Proposed interface:

```py
orders_table = DynamoTable(
    name="orders",
    fields={
        "customer_id": "string",
        "order_date": "string",
    },
    partition_key="customer_id",
    sort_key="order_date",
    stream="new-and-old-images",
)

orders_table.subcribe("order-changes", handler="functions/orders_events.handler")
```

michal-stlv:
Closes #38

### DIFF
--- a/stelvio/aws/function/__init__.py
+++ b/stelvio/aws/function/__init__.py
@@ -11,7 +11,7 @@ from pulumi_aws.iam import GetPolicyDocumentStatementArgs, Policy, Role
 
 from stelvio import context
 from stelvio.aws.permission import AwsPermission
-from stelvio.component import Component
+from stelvio.component import Component, safe_name
 from stelvio.link import Link, Linkable
 from stelvio.project import get_project_root
 
@@ -137,7 +137,7 @@ class Function(Component[FunctionResources]):
         function_runtime = self.config.runtime or DEFAULT_RUNTIME
         function_architecture = self.config.architecture or DEFAULT_ARCHITECTURE
         function_resource = lambda_.Function(
-            context().prefix(self.name),
+            safe_name(context().prefix(), self.name, 64),
             role=lambda_role.arn,
             architectures=[function_architecture],
             runtime=function_runtime,
--- a/tests/aws/api_gateway/test_api.py
+++ b/tests/aws/api_gateway/test_api.py
@@ -1,14 +1,15 @@
 import json
-import re
 from dataclasses import dataclass, field, replace
 from pathlib import Path
 from typing import Literal, cast
+from unittest.mock import patch
 
 import pulumi
 import pytest
 from pulumi import StringAsset
 from pulumi.runtime import set_mocks
 
+from stelvio import context
 from stelvio.aws.api_gateway import API_GATEWAY_ROLE_NAME, Api, ApiConfig
 from stelvio.aws.function import Function, FunctionAssetsRegistry, FunctionConfig
 from stelvio.component import ComponentRegistry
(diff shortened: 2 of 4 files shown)


sebst:
What's the reason to support both `StreamView`, and `StreamViewLiteral`?

michal-stlv:
> What's the reason to support both `StreamView`, and `StreamViewLiteral`?

Giving user options. Some might prefer to use enums (for different reasons) but some might prefer just to use string literal. This is common pattern in stelvio with most of the options.

sebst (approved):

docs/guides/dynamo-db.md
```diff
@@ -124,6 +124,69 @@ LocalIndex(sort_key="status", projections=["total", "customer_id"])
     - **Global indexes** can be added or removed from existing tables, but this is a background operation that takes time  
     - Each table can have up to 5 local indexes and 20 global indexes
 
+## DynamoDB Streams
+
+DynamoDB Streams capture changes to your table data in real-time. Enable streams to react to creates, updates, and deletes:
```
sebst, line 141: Enable

docs/guides/dynamo-db.md
```diff
@@ -124,6 +124,69 @@ LocalIndex(sort_key="status", projections=["total", "customer_id"])
...
+
+```python
+# Check if streams are enabled
+if orders_table._config.stream_enabled:
+    stream_arn = orders_table.stream_arn  # Output[str] | None
+    # Use stream_arn in other Pulumi resources that need it
+```
+
+!!! warning "Stream Retention"
+    DynamoDB streams retain data for **24 hours only**. After that, the data is automatically deleted. Plan your stream processing accordingly.
```
sebst: Consider explaining that this is NOT configurable

docs/guides/dynamo-db.md
michalmartinka: There is Enable no?

stelvio/aws/dynamo_db.py
```diff
@@ -189,10 +347,60 @@ def _create_resources(self) -> DynamoTableResources:
...
+            # Create function with merged permissions
+            function_name = f"{self.name}-{subscription.name}"
+            function = Function(function_name, config_with_stream)
+
+            # Create EventSourceMapping
+            mapping = EventSourceMapping(
+                context().prefix(f"{self.name}-{subscription.name}-mapping"),
+                event_source_arn=table.stream_arn,
+                function_name=function.function_name,
+                starting_position="LATEST",
```
sebst: can the user change `starting_position`?

michalmartinka: no, we could add it but i think it's so rare that we'd just leave it for transform (next version). However i'm now implementing support for batch_size and filters

sebst: Same for batch_size and maximum_batching_window_in_seconds

docs/guides/dynamo-db.md
sebst: Yes. I swear I read "Enables" lol

stelvio/aws/dynamo_db.py
```diff
@@ -319,8 +319,10 @@ def _create_subscription(
         opts: FunctionConfigDict,
     ) -> _DynamoSubscription:
         subscription_config = (
-            config if isinstance(config, SubscriptionConfig)
-            else SubscriptionConfig(**config) if config
+            config
+            if isinstance(config, SubscriptionConfig)
+            else SubscriptionConfig(**config)
+            if config
```
sebst: That took me a while ;)

sebst (dismissed):

stelvio/aws/dynamo_db.py
michal-stlv: you're right that's too much. changed to
```py
        if config is None:
            subscription_config = SubscriptionConfig()
        elif isinstance(config, SubscriptionConfig):
            subscription_config = config
        else:
            subscription_config = SubscriptionConfig(**config)
```

sebst (approved):

stelvio/aws/dynamo_db.py
```diff
@@ -187,28 +187,135 @@ def __post_init__(self) -> None:
 
 @dataclass(frozen=True)
 class SubscriptionConfig:
-    filters: dict | None = None
+    filters: list[dict] | None = None
```
sebst, line 190: Can we have a typedict?

stelvio/aws/dynamo_db.py
```diff
@@ -187,28 +187,135 @@ def __post_init__(self) -> None:
...
+
+        # Create new config with merged links
+        config_with_merged_links = replace(self.handler, links=merged_links)
+
+        # Create function with merged permissions
+        function = Function(self.function_name, config_with_merged_links)
+
+        # Create EventSourceMapping - table.stream_arn triggers table creation naturally
+        mapping = EventSourceMapping(
+            context().prefix(f"{self.name}-mapping"),
```
sebst, line 286: safe_name?

stelvio/aws/dynamo_db.py
michal-stlv: theoretically maybe . practically i don't think we want to do it because possible filters are quite broad . See https://docs.aws.amazon.com/lambda/latest/dg/invocation-eventfiltering.html#filtering-syntax

it would be interesting to investigate what we could do here to help people write filters. wheter it's typed dicts or some other way. But I think not in the scope of this PR or issue.

stelvio/aws/dynamo_db.py
michal-stlv: i did some research and it shouldn't be needed here. that name is not used by aws, only by pulumi. aws uses uuid for EventSourceMapping. 
However later  in `_create_resources` we create function (lambda) and it has hard limit on name, so if subscription name with prefix will be longer deployment will fail. Also I've just noticed same is case for function created automatically by `route()` in `Api`. I guess we need to bake `safe_name` directly into `Function` although it will also truncate user defined functions. What do you think?

stelvio/aws/dynamo_db.py
sebst: So, we leave it for now. We should investigate if we can generate python type files from terraform, pulumi or openapi specs.

stelvio/aws/dynamo_db.py
michal-stlv: ok we're now using safe_name inside Function

PR #41, opened 2025-09-13
Static Website Support for S3

### LINKED ISSUE #32: Enhanced S3 support: Static Website and Cloudfront
To extend the S3 support in #23, we should add support for static websites using cloudfront distributions and a dedicated static website component. (see PR #25) for general ideas.

sebst:
Closes #32

### DIFF
--- a/stelvio/aws/cloudfront.py
+++ b/stelvio/aws/cloudfront.py
@@ -76,10 +76,9 @@ class CloudFrontDistribution(Component[CloudFrontDistributionResources]):
         )
 
         # Create CloudFront Distribution
-        aliases = [self.custom_domain] if self.custom_domain else None
         distribution = pulumi_aws.cloudfront.Distribution(
             context().prefix(self.name),
-            aliases=aliases,
+            aliases=[self.custom_domain] if self.custom_domain else None,
             origins=[
                 {
                     "domain_name": self.bucket.resources.bucket.bucket_regional_domain_name,


stelvio/aws/s3.py
```diff
@@ -110,3 +115,115 @@ def default_bucket_link(bucket: pulumi_aws.s3.Bucket) -> LinkConfig:
...
+@dataclass(frozen=True)
+class S3StaticWebsiteResources:
+    bucket: pulumi_aws.s3.Bucket
+    files: list[pulumi_aws.s3.BucketObject]
+    cloudfront_distribution: CloudFrontDistribution
+
+
+@final
+class S3StaticWebsite(Component[S3StaticWebsiteResources]):
+    def __init__(self, name: str, custom_domain: str, directory: Path | str | None = None):
```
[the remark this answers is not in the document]
sebst: This is reflected in the documentation

docs/guides/s3.md
```diff
@@ -54,20 +54,100 @@ def run() -> None:
...
+    _ = S3StaticWebsite(
+        "s3-static-mkdocs",
+        custom_domain="s3-2." + CUSTOM_DOMAIN_NAME,
+        directory="site"
+    )
+```
+
+- Creates an S3 bucket for static website hosting
+- Creates a CloudFront distribution for the S3 bucket, so that it is compatible with 3rd party DNS providers
+  - If CloudFront distribution is omitted, a DNS record will need to be created with the same name as the S3 bucket which is obviously not possible with 3rd party DNS providers
```
michal-stlv: I think this might need more explanation? why it's not possible?

docs/guides/s3.md
```diff
@@ -54,20 +54,100 @@ def run() -> None:
...
+    s3_bucket = website.bucket
+    boto3_client = boto3.client("s3")
+    boto3_client.put_object(
+        Bucket=s3_bucket.bucket_name, Key="index.html", Body="<h1>Hello, World!</h1>"
+    )
+```
+
+In both cases, the files uploaded to your static website are considered part of your infrastructure. Thus, the files would be automatically deployed whenever you run `stlv deploy`.
+
+Using the `stlv_resources` module, you can access the S3 bucket and manage the files (assets) in your static website from anywhere in your project, should you decide that part should not be part of your deployment.
```
michal-stlv: I know what you mean here but not sure it's clear to new users. "anywhere" is too broad. basically now you can get bucket arn from `Resources` class in `stlv_resources` module in your lambdas only (if created with Stelvio's `Function`)

docs/guides/s3.md
```diff
@@ -54,20 +54,100 @@ def run() -> None:
...
+
+### Parameters
+
+| Parameter       | Description                                                                                                                                            |
+| --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `custom_domain` | The custom domain name for the static website. Optional. If provided, a DNS record will be created for the CloudFront distribution. Optional. A `str`. |
```
michal-stlv, line 157: what if you dont' provide custom domain? what will happen? how do i access my public bucket?

stelvio/aws/cloudfront.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Update S3 bucket policy to allow CloudFront access
+        bucket_policy = pulumi_aws.s3.BucketPolicy(
+            context().prefix(f"{self.name}-bucket-policy"),
+            bucket=self.s3_bucket.id,
+            policy=pulumi.Output.all(
+                distribution_arn=distribution.arn,
+                bucket_arn=self.s3_bucket.arn,
+            ).apply(
+                lambda args: pulumi.Output.json_dumps(
```
michal-stlv, line 147: i think you should be able to unpack args an use `lambda (distribution_arn, bucket_arn): pulumi.Output.json_dumps(` or similar syntax so you dont' have to do `args['bucket_arn']` in the body of the lambda.

stelvio/aws/cloudfront.py
```diff
@@ -0,0 +1,168 @@
...
+    distribution: pulumi_aws.cloudfront.Distribution
+    origin_access_control: pulumi_aws.cloudfront.OriginAccessControl
+    acm_validated_domain: AcmValidatedDomain
+    record: Record
+    bucket_policy: pulumi_aws.s3.BucketPolicy
+    function_associations: list[dict]
+
+
+@final
+class CloudFrontDistribution(Component[CloudFrontDistributionResources]):
```
michal-stlv, line 40: Is `CloudFrontDistribution` supposed to be used separately, outside of `S3StaticWebsite` component? 

If NO: I would make it `_ CloudFrontDistribution` for now? and also wouldn't mention it in in changelogs or docs really.

If YES: how? is it documented how it make sense without S3StaticWebsite component? Also in that case `s3_bucket` should be Stelvio's `Bucket` component instead of Pulumi's. Generally in stelvio interfaces we should use stelvio components if possible rather than Pulumi's.

stelvio/aws/s3.py
```diff
@@ -110,3 +115,116 @@ def default_bucket_link(bucket: pulumi_aws.s3.Bucket) -> LinkConfig:
...
+                    if (uri.endsWith('/')) {
+                        request.uri += 'index.html';
+                    }
+                    // Check whether the URI is missing a file extension.
+                    else if (!uri.includes('.')) {
+                        request.uri += '/index.html';
+                    }
+                    return request;
+                }
+            """.strip(),
```
michal-stlv: Maybe worth extracting it into constant outside of the method?

stelvio/aws/s3.py
```diff
@@ -110,3 +115,116 @@ def default_bucket_link(bucket: pulumi_aws.s3.Bucket) -> LinkConfig:
...
+
+                    bucket_object = pulumi_aws.s3.BucketObject(
+                        resource_name,
+                        bucket=bucket.resources.bucket.id,
+                        key=str(key),
+                        source=pulumi.FileAsset(file_path),
+                        content_type=mimetype,
+                        cache_control=cache_control,
+                    )
+                    files.append(bucket_object)
```
michal-stlv: This could be separate method or even function and maybe try to do nice comprehension instead of nested loops?

stelvio/aws/cloudfront.py
```diff
@@ -0,0 +1,168 @@
...
+                depends_on=[distribution]
+            ),  # Ensure policy is applied after distribution
+        )
+
+        record = context().dns.create_record(
+            resource_name=f"{self.name}-cloudfront-record",
+            name=self.custom_domain,
+            record_type="CNAME",
+            value=distribution.domain_name,
+            ttl=1,
```
michal-stlv: `ttl=1` are we supposed to set it to 1?

stelvio/aws/cloudfront.py
```diff
@@ -0,0 +1,168 @@
...
+
+@final
+class CloudFrontDistribution(Component[CloudFrontDistributionResources]):
+    def __init__(
+        self,
+        name: str,
+        s3_bucket: pulumi_aws.s3.Bucket,
+        custom_domain: str,
+        price_class: CloudfrontPriceClass = "PriceClass_100",
+        function_associations: list[dict] | None = None,
```
michal-stlv: can we do better than just `list[dict]`? do we know format of the dict? maybe we can define some typeddict? so user gets IDE support or can look at type to know format?

tests/aws/s3/test_cloudfront.py
```diff
@@ -0,0 +1,411 @@
...
+    @property
+    def type(self):
+        return self.pulumi_resource.type
+
+    @property
+    def value(self):
+        return self.pulumi_resource.content
+
+
+class MockDns(Dns):
```
michal-stlv: we already have MockDns, even two of them?

tests/aws/s3/test_cloudfront.py
```diff
@@ -0,0 +1,411 @@
...
+            dns=mock_dns,
+        )
+    )
+    yield mock_dns
+    _ContextStore.clear()
+
+
+@pytest.fixture
+def app_context_without_dns():
+    """App context without DNS provider configured"""
```
michal-stlv: I'm all for comments and docstrings but only if they add any extra info or context. here `"""App context without DNS provider configured"""` doesn't add anything to function called `app_context_without_dns`, does it? Same for `app_context_with_dns`

stelvio/aws/s3.py
```diff
@@ -110,3 +115,116 @@ def default_bucket_link(bucket: pulumi_aws.s3.Bucket) -> LinkConfig:
             ),
         ],
     )
+
+
+@dataclass(frozen=True)
```
michal-stlv: final?

stelvio/aws/s3.py
```diff
@@ -110,3 +115,116 @@ def default_bucket_link(bucket: pulumi_aws.s3.Bucket) -> LinkConfig:
...
+class S3StaticWebsiteResources:
+    bucket: pulumi_aws.s3.Bucket
+    files: list[pulumi_aws.s3.BucketObject]
+    cloudfront_distribution: CloudFrontDistribution
+
+
+@final
+class S3StaticWebsite(Component[S3StaticWebsiteResources]):
```
michal-stlv: I think `S3StaticWebsite` should be in separate file. What do you think. In the end it's separate component.  `s3_static_website.py`?

michal-stlv (commented):
Great stuff Bas overall. Let's polish the details and it will be good to go

stelvio/aws/cloudfront.py
sebst: We did it for `acm` and `ApiGateway`, too. 
Probably not optimal, but this way we ensure that a `stlv deploy` is effective as early as possible

stelvio/aws/s3.py
sebst: > I think `S3StaticWebsite` should be in separate file. What do you think. In the end it's separate component. `s3_static_website.py`?

Just from the user's pov, I'd prefer to 
`import S3StaticWebsite from stelvio.aws.s3`
rather than
`import S3StaticWebsite from stelvio.aws.s3_static_website`

On the other hand, I'd prefer a separate file, too (from a maintainability pov).

I'd suggest to move `s3.py` to `s3/s3.py` (with a proper `__init__.py`) and move `S3StaticWebsite` to `s3/static_website.py`. We could generalize it:
- Directories (modules) for AWS services (`s3`, `cloudfront`, etc)
- Files for each component (and their internally used helper classes/functions/constants)

stelvio/aws/s3.py
sebst: wdym?

stelvio/aws/cloudfront.py
sebst: > Is `CloudFrontDistribution` supposed to be used separately, outside of `S3StaticWebsite` component?

Yes, but not at the moment. The reason, it might be re-used is what made me extract the `function_associations`.
It is still tighly coupled to the `S3StaticWebsite` for the time being, so right now, it's not supposed to be used outside of it.

As a core AWS service, we should therefore investigate further how people use it in real life to make it re-usable for those folks.

> If YES: how? is it documented how it make sense without S3StaticWebsite component? Also in that case `s3_bucket` should be Stelvio's `Bucket` component instead of Pulumi's. Generally in stelvio interfaces we should use stelvio components if possible rather than Pulumi's.

No mention in the docs because as outlined, not really useful for anything other than static websites. Will add the doc with the next update on cloudfront.

Will change the coponent vs resource type, though.

stelvio/aws/cloudfront.py
sebst: > i think you should be able to unpack args an use `lambda (distribution_arn, bucket_arn): pulumi.Output.json_dumps(` or similar syntax so you dont' have to do `args['bucket_arn']` in the body of the lambda.

Could you please have a look at [this commit](https://github.com/stelviodev/stelvio/pull/41/commits/17cfefac0d8e2214235dcaba323c80913ad8fadf) -- it seems to create a ruff check error....

stelvio/aws/s3.py
michalmartinka: we can use `__all__` so user can import from s3 or even aws namespace, this is something to review and do anyway.

stelvio/aws/s3.py
michalmartinka: sorry i meant if datacalss should be final 
```py
@final
@dataclass(frozen=True)
```

stelvio/aws/cloudfront.py
michal-stlv: Ok, I understand and agree we should (later) support `CloudFrontDistribution` as a standalone component and it has many use cases. But right now it's tightly couple with `S3StaticWebsite` therefore I'd not mention it in changelog or anywhere and make it private/protected using `_CloudFrontDistribution`.

stelvio/aws/s3/s3_static_website.py
```diff
@@ -0,0 +1,137 @@
...
+class S3StaticWebsite(Component[S3StaticWebsiteResources]):
+    def __init__(self, name: str, custom_domain: str, directory: Path | str | None = None):
+        super().__init__(name)
+        self.directory = Path(directory) if isinstance(directory, str) else directory
+        self.custom_domain = custom_domain
+        self._resources = None
+
+    def _create_resources(self) -> S3StaticWebsiteResources:
+        # Validate directory exists
+        if self.directory is None or not self.directory.exists():
```
michal-stlv: We don't want to raise if `self.directory is None` do we ? Shouldn't test catch this mistake? Or am I missing something?

stelvio/aws/s3/s3_static_website.py
```diff
@@ -0,0 +1,137 @@
...
+            function_associations=[
+                {
+                    "event_type": "viewer-request",
+                    "function_arn": viewer_request_function.arn,
+                }
+            ],
+        )
+
+        # Upload files from directory to S3 bucket
+        files = self._process_directory_and_upload_files(bucket, self.directory)
```
michal-stlv, line 82: Related to my comment about `if self.directory is None` we shoud not do this always only if user provided directory right?

michal-stlv: nevermind i see you check inside `_process_directory_and_upload_files`

stelvio/aws/s3/s3_static_website.py
michal-stlv: Or did you mean `if self.directory is not None and not self.directory.exists():` ? but still tests should catch this bug.

stelvio/aws/s3/s3_static_website.py
```diff
@@ -0,0 +1,137 @@
...
+                    bucket_object = pulumi_aws.s3.BucketObject(
+                        resource_name,
+                        bucket=bucket.resources.bucket.id,
+                        key=str(key),
+                        source=pulumi.FileAsset(file_path),
+                        content_type=mimetype,
+                        cache_control=cache_control,
+                    )
+                    files.append(bucket_object)
+        return files
```
michal-stlv: I meant to extract creation of `BucketObject` into separatre funciton and then use it in comprehension. e.g. (but please keep all your comments from original):
```py
def _create_s3_bucket_object(
    self, bucket: Bucket, directory: Path, file_path: Path
) -> pulumi_aws.s3.BucketObject:
    key = file_path.relative_to(directory)

    # Convert path separators and special chars to dashes,
    # ensure valid Pulumi resource name
    safe_key = re.sub(r"[^a-zA-Z0-9]", "-", str(key))
    # Remove consecutive dashes and leading/trailing dashes
    safe_key = re.sub(r"-+", "-", safe_key).strip("-")
    # resource_name = f"{self.name}-{safe_key}-{file_hash[:8]}"

    # DO NOT INCLUDE HASH IN RESOURCE NAME
    # If the resource name changes, Pulumi will treat it as a new resource,
    # and create a new s3 object
    # Then, the old one is deleted by pulumi. Sounds correct, but since the
    # filename (key) is the same, the delete operation deletes the new object!
    resource_name = f"{self.name}-{safe_key}"

    # For binary files, use source instead of content
    mimetype, _ = mimetypes.guess_type(file_path.name)

    cache_control = "public, max-age=1"  # 1 second

    return pulumi_aws.s3.BucketObject(
        resource_name,
        bucket=bucket.resources.bucket.id,
        key=str(key),
        source=pulumi.FileAsset(file_path),
        content_type=mimetype,
        cache_control=cache_control,
    )
```

and then 
```py
def _process_directory_and_upload_files(
    self, bucket: Bucket, directory: Path
) -> list[pulumi_aws.s3.BucketObject]:
    # glob all files in the directory
    if directory is None:
        return []

    return [
        self._create_s3_bucket_object(bucket, directory, Path(root) / filename)
        for root, _, filenames in os.walk(directory)
        for filename in filenames
    ]
```

OR we could use pathlib instead:
```py
def _process_directory_and_upload_files(
    self, bucket: Bucket, directory: Path
) -> list[pulumi_aws.s3.BucketObject]:
    # glob all files in the directory
    if directory is None:
        return []

    return [
        self._create_s3_bucket_object(bucket, directory, file_path)
        for file_path in directory.rglob("*") if file_path.is_file()
    ]
```

stelvio/aws/s3/s3_static_website.py
```diff
@@ -0,0 +1,137 @@
...
+                    # Remove consecutive dashes and leading/trailing dashes
+                    safe_key = re.sub(r"-+", "-", safe_key).strip("-")
+                    # resource_name = f"{self.name}-{safe_key}-{file_hash[:8]}"
+
+                    # DO NOT INCLUDE HASH IN RESOURCE NAME
+                    # If the resource name changes, Pulumi will treat it as a new resource,
+                    # and create a new s3 object
+                    # Then, the old one is deleted by pulumi. Sounds correct, but since the
+                    # filename (key) is the same, the delete operation deletes the new object!
+                    resource_name = f"{self.name}-{safe_key}"
```
michal-stlv: Also here or on line 129 I think we should use `prefix` and also `safe_name` .

stelvio/aws/s3.py
michal-stlv: Sorry I meant it could be 
```py
@final
@dataclass(frozen=True)
```

stelvio/aws/cloudfront.py
michal-stlv: I looked at the commit but i'm not sure what you're trying to do there, sorry. I thought we could unpack args into multiple params but that doesn't work. I was after slight improvement but if Ruff (rightly) complains with two lambdas,  maybe let's just use your original code to keep it simple, ok? Sorry again for wasted effort.

stelvio/aws/cloudfront.py
michal-stlv: Yeah, we did it. And it works, and it's helpful during development as it makes sure stuff is updated/refreshed in DNS immediately.

BUT 

It's not a good practice for production environments because it basically means that records are not cached which means that then every time DNS is hit it can't use cache and needs to lookup record which also increases the cost of Route53.

So I suggest to use `300` and see what happens. Or ideally we'd use `300` for prod and 1 for others BUT we dont' know how users will name their prod envs.  Might be `prod`, might be `production`, might be `live`, whatever. So maybe we need to think about this a bit. Or use some middle ground value in between for now until we figure it out but in that case please put `# TODO: ` in code and create issue.  But I'm afraid `ttl=1` is not good for Stelvio production use which is where we are heading.

docs/changelog.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.5.0a1 (2025-09-13)
+
+With this release, Stelvio gets a StaticWebsite component for S3 static website hosting with CloudFront CDN and optional custom domain support.
+
+### Static Website Hosting with S3 and CloudFront
+- Added `stelvio.aws.s3.S3StaticWebsite` for managing S3 buckets for static website hosting with CloudFront CDN and optional custom domain support
+- Added `stelvio.aws.cloudfront.CloudFrontDistribution` for managing CloudFront distributions
```
michal-stlv: As discussed in other comment, if `stelvio.aws.cloudfront.CloudFrontDistribution` can't be used separately I'd not mention it here

docs/guides/s3.md
```diff
@@ -54,20 +54,103 @@ def run() -> None:
...
+
+    # Upload files to the bucket
+    s3_bucket = website.bucket
+    boto3_client = boto3.client("s3")
+    boto3_client.put_object(
+        Bucket=s3_bucket.bucket_name, Key="index.html", Body="<h1>Hello, World!</h1>"
+    )
+```
+
+In both cases, the files uploaded to your static website are considered part of your infrastructure. Thus, the files would be automatically deployed whenever you run `stlv deploy`.
```
michal-stlv: > files uploaded to your static website are considered part of your infrastructure

I know what you mean, yes. But there's a distinction. If you use directory file are also part of your state. But if you have your own code that uploads them then it's not part of the state which has implications regarding e.g. deletion of the file etc. Maybe worth mentioning or explaining .

stelvio/aws/cloudfront.py
```diff
@@ -0,0 +1,178 @@
...
+
+from stelvio import context
+from stelvio.aws.acm import AcmValidatedDomain
+from stelvio.component import Component
+
+if TYPE_CHECKING:
+    from stelvio.aws.s3.s3 import Bucket
+    from stelvio.dns import Record
+
+CloudfrontPriceClass = Literal["PriceClass_100", "PriceClass_200", "PriceClass_All"]
```
michal-stlv, line 21: Oh, I really dont' like names of these price classes. I know that's what aws calles them but man, it's ugly :) I don't have any idea here right now but maybe let's put `# TODO: ` to think about it?

stelvio/aws/s3.py
sebst: We don't have that on any other `*Resources` class. Should we? If so, then we'd add it to all other `*Resources` classes in another ticket...

docs/changelog.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.5.0a1 (2025-09-13)
```
michal-stlv: this needs to be `a7`, this number (alpha version) doesn't reset with version. but we can fix this before release.

stelvio/aws/cloudfront.py
```diff
@@ -0,0 +1,181 @@
...
+                    }
+                )
+            ),
+            opts=pulumi.ResourceOptions(
+                depends_on=[distribution]
+            ),  # Ensure policy is applied after distribution
+        )
+
+        record = context().dns.create_record(
+            resource_name=f"{self.name}-cloudfront-record",
```
michal-stlv: Don't we need to use `context().prefix` here? and potentially `safe_name()` if there's limit in aws for this resoruce

stelvio/aws/cloudfront.py
```diff
@@ -0,0 +1,181 @@
...
+                        ],
+                    }
+                )
+            ),
+            opts=pulumi.ResourceOptions(
+                depends_on=[distribution]
+            ),  # Ensure policy is applied after distribution
+        )
+
+        record = context().dns.create_record(
```
michal-stlv: will this fail if `context().dns` is not set ?

stelvio/aws/s3/s3_static_website.py
```diff
@@ -0,0 +1,140 @@
...
+        request.uri += '/index.html';
+    }
+    return request;
+}
+"""
+
+
+@final
+class S3StaticWebsite(Component[S3StaticWebsiteResources]):
+    def __init__(self, name: str, custom_domain: str, directory: Path | str | None = None):
```
michal-stlv: isn't `custom_domain` optional?

stelvio/aws/cloudfront.py
```diff
@@ -0,0 +1,181 @@
...
+            viewer_certificate={
+                "acm_certificate_arn": acm_validated_domain.resources.certificate.arn,
+                "ssl_support_method": "sni-only",
+                "minimum_protocol_version": "TLSv1.2_2021",
+            },
+            custom_error_responses=[
+                {
+                    "error_code": 403,
+                    "response_code": 404,
+                    "response_page_path": "/error.html",
```
michal-stlv, line 127: all of this assumes `/error.html` exists, which i guess might be ok if it's documented but it's not? will it enter some 404 error loop? (for user)

stelvio/aws/s3/s3_static_website.py
```diff
@@ -0,0 +1,140 @@
...
+        # If the resource name changes, Pulumi will treat it as a new resource,
+        # and create a new s3 object
+        # Then, the old one is deleted by pulumi. Sounds correct, but since the
+        # filename (key) is the same, the delete operation deletes the new object!
+        resource_name = f"{self.name}-{safe_key}"
+
+        # For binary files, use source instead of content
+        mimetype, _ = mimetypes.guess_type(file_path.name)
+
+        cache_control = "public, max-age=1"  # 1 second
```
michal-stlv: 1 second for cache? doesn't that make any cache useless? maybe settings? or some more reasonable value? I guess even for dev env it should be more and for prod much more. Otherwise cloudfront is not useful here and s3 bucket will get hit every time?

stelvio/aws/s3/s3_static_website.py
```diff
@@ -0,0 +1,140 @@
...
+            raise FileNotFoundError(f"Directory does not exist: {self.directory}")
+
+        bucket = Bucket(f"{self.name}-bucket")
+        # Create CloudFront Function to handle directory index rewriting
+        viewer_request_function = pulumi_aws.cloudfront.Function(
+            context().prefix(f"{self.name}-viewer-request"),
+            name=context().prefix(f"{self.name}-viewer-request-function"),
+            runtime="cloudfront-js-1.0",
+            comment="Rewrite requests to directories to serve index.html",
+            code=REQUEST_INDEX_HTML_FUNCTION_JS,
```
michal-stlv: This is nice but maybe some TODO somehwere to note that e.g. if in the future we'd use this component for SPA apps we might need to add common SPA routing patterns or make this configurable.

docs/guides/s3.md
```diff
@@ -54,20 +54,107 @@ def run() -> None:
...
-Future releases of Stelvio will allow public access to S3 buckets, like static websites, using Cloudfront. This is still in development.
\ No newline at end of file
+Stelvio can create and manage S3 buckets for static website hosting using the `S3StaticWebsite` component.
+
+Create a static website from a directory using the `S3StaticWebsite` component:
+
+```python
+@app.run
+def run() -> None:
+    config = mkdocs.config.load_config("mkdocs.yml")
```
michal-stlv, line 95: maybe small explanation that mkdcos here is just example and not requirement.

stelvio/aws/cloudfront.py
```diff
@@ -0,0 +1,181 @@
...
+                {
+                    "domain_name": self.bucket.resources.bucket.bucket_regional_domain_name,
+                    "origin_id": f"{self.name}-S3-Origin",
+                    "origin_access_control_id": origin_access_control.id,
+                }
+            ],
+            enabled=True,
+            is_ipv6_enabled=True,
+            default_root_object="index.html",
+            default_cache_behavior={
```
michal-stlv, line 92: One thing I'm wondering about - should we also add security headers configuration (HSTS, CSP, etc.) to the CloudFront distribution? I'm not an expert on this (yet), but I noticed AWS has some guidance on it: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-response-headers-policies.html

Not necessarily for this PR, but thought it might be worth considering for the default setup or as an optional parameter in a future iteration. (Maybe leave a todo or we can create issue if we agree on this). 

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/modifying-response-headers.html

michal-stlv (commented):
Shou

stelvio/aws/cloudfront.py
sebst: #43

stelvio/aws/cloudfront.py
```diff
@@ -68,9 +76,10 @@ def _create_resources(self) -> CloudFrontDistributionResources:
         )
 
         # Create CloudFront Distribution
+        aliases = [self.custom_domain] if self.custom_domain else None
         distribution = pulumi_aws.cloudfront.Distribution(
             context().prefix(self.name),
-            aliases=[self.custom_domain],
+            aliases=aliases,
```
michal-stlv: we can inline `aliases` ?  
```py
aliases=[self.custom_domain] if self.custom_domain else None,
```

tests/aws/s3/test_static_website.py
```diff
@@ -157,15 +157,15 @@ def test_s3_static_website_without_custom_domain(
...
         website = S3StaticWebsite(
             name="test-website-no-domain",
             directory=str(temp_static_site),
             custom_domain="www.example.com",  # This will fail without DNS
         )
-
-        # This should raise an error when trying to create ACM certificate
         _ = website.resources
+    # Assert the error message is about missing DNS
+    assert "dns" in str(exc_info.value).lower() or "DNS" in str(exc_info.value)
```
michal-stlv, line 168: what are we checking here? jsut that `"dns"` is in  error message ? and is it `"dns"` or `"DNS"` ?

sebst: Yeah, that's basically just checking the message

michal-stlv (approved):
Great stuff. Let's merge.

PR #60, opened 2025-10-26
Add npm package src

### LINKED ISSUE #59: Add wrapper package on npmjs for easier onboarding
[https://www.npmjs.com/package/stlv](https://www.npmjs.com/package/stlv)

sebst:
Closes #59

### DIFF
--- a/pkg/npm/LICENSE
+++ b/pkg/npm/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this wdocument.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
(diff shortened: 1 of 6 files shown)


pkg/npm/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import { execSync } from "child_process";
+
+const ensureUv = () => {
+	// check if 'uv' is a valid executable in the PATH
+	// then run 'uv --version' to verify it's working
+	// if not found, throw an error
+	try {
+		const version = execSync("uv --version", { encoding: "utf-8" });
+		console.log(`uv is installed. Version: ${version}`);
```
michal-stlv, line 11: These logs print every time someone runs a command? maybe lot of noise? should print only in --verbose?

pkg/npm/index.ts
```diff
@@ -0,0 +1,59 @@
...
+	} catch (error) {
+		throw new Error("uv executable not found in PATH. Please install uv.");
+	}
+};
+
+const ensureUvx = () => {
```
michal-stlv, line 17: Why we check for av and uvx? its all in one package? when uv is installed uvx is too?

pkg/npm/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const ensureStelvio = () => {
+	// check if 'stelvio' is a valid executable in the PATH
+	// then run 'stelvio --version' to verify it's working
+	// if not found, throw an error
+	try {
+		execSync("uvx --from stelvio stlv --help", { encoding: "utf-8" });
+	} catch (error) {
+		throw new Error(
+			"stelvio executable not found in PATH. Please install stelvio.",
```
michal-stlv, line 37: It's not rally in path is it? shouldn't message be more like "Failed to run stelvio. Make sure uv is properly installed."

pkg/npm/index.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+const runStelvioCommand = (args: string[]) => {
+	try {
+		const command = `uvx --from stelvio stlv ${args.join(" ")}`;
+		const output = execSync(command, { encoding: "utf-8" });
+		console.log(output);
+	} catch (error) {
+		console.error("Error running stelvio command:", error);
+	}
```
michal-stlv, line 49: When stelvio command fails, the npm wrapper exits with code 0 (success)? 

```py
catch (error: any) {
    // Propagate exit code
    process.exit(error.status || 1);
      }
```

pkg/npm/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this wdocument.
```
michal-stlv, line 10: wdocument -> document ?

pkg/npm/package.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/stelviodev/stelvio/issues"
+  },
+  "bin": {
+    "stlv": "./dist/index.js"
+  },
+  "devDependencies": {
+    "@types/node": "^20",
+    "typescript": "^5"
```
michal-stlv, line 22: Uses ^20 for @types/node but the bun.lock has node 24?

pkg/npm/index.ts
```diff
@@ -0,0 +1,59 @@
...
+			"stelvio executable not found in PATH. Please install stelvio.",
+		);
+	}
+};
+
+const runStelvioCommand = (args: string[]) => {
+	try {
+		const command = `uvx --from stelvio stlv ${args.join(" ")}`;
+		const output = execSync(command, { encoding: "utf-8" });
+		console.log(output);
```
michal-stlv, line 46: ## Interactive commands don't work

**Problem:** Using `{ encoding: "utf-8" }` at line 45 captures output instead of streaming it. This breaks:
- Interactive prompts (e.g., `stlv deploy production` confirmation won't show and user can't respond)
- Real-time progress output (nothing prints until command finishes)
- Color formatting

**Current code:**
```typescript
const output = execSync(command, { encoding: "utf-8" });
console.log(output);
```

**Fix:**
```typescript
execSync(command, { stdio: "inherit" });
```

`stdio: "inherit"` passes stdin/stdout/stderr directly through to the child process, making it behave like running stelvio directly.

PR #64, opened 2025-10-27
Improve templates

sebst:
Add unit test
Add split repo
Allow for thirdparty repos

### DIFF
--- a/stelvio/cli/__init__.py
+++ b/stelvio/cli/__init__.py
@@ -1,6 +1,7 @@
 import getpass
 import logging
 import os
+import sys
 from collections.abc import Callable
 from importlib import metadata
 from logging.handlers import TimedRotatingFileHandler
@@ -61,11 +62,21 @@ def safe_run_pulumi(func: Callable, env: str | None, **kwargs: bool) -> None:
         raise click.Abort from e
 
 
-@click.group()
+@click.group(invoke_without_command=True)
 @click.option(
     "--verbose", "-v", count=True, help="Increase verbosity. -v for INFO, -vv for DEBUG logs."
 )
-def cli(verbose: int) -> None:
+@click.option("--version", is_flag=True, help="Show Stelvio and Pulumi versions.")
+@click.pass_context
+def cli(ctx: click.Context, verbose: int, version: bool) -> None:
+    if version:
+        _version()  # Exits after printing versions
+
+    # If no command was invoked, show help
+    if ctx.invoked_subcommand is None:
+        console.print(ctx.get_help())
+        ctx.exit(0)
+
     if verbose > 0:
         console_handler = RichHandler(
             console=console,
@@ -85,10 +96,6 @@ def cli(verbose: int) -> None:
         console.print(f"[italic dim]Logs saved to: {log_file_path}[/]")
         app_logger.addHandler(console_handler)
 
-    if needs_pulumi():
-        with console.status("Downloading Pulumi..."):
-            install_pulumi()
-
 
 @click.command()
 @click.option("--template", default=None, help="Template to use for initialization")
@@ -97,6 +104,7 @@ def init(template: str | None) -> None:
     Initialize a Stelvio project in the current directory.
     Creates stlv_app.py with AWS configuration template.
     """
+    _ensure_pulumi()
     stelvio_art(console)
     stlv_app_path, app_exists = get_stlv_app_path()
     if app_exists:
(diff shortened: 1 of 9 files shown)


sebst:
Would need a PAT for this, consider before merging! (see Action)

stelvio/cli/__init__.py
```diff
@@ -249,3 +250,57 @@ def determine_env(environment: str) -> str:
...
+    # owner: owner, repo: repo, branch: branch, subdirectory: None
+
+    # gh:owner/repo/subdirectory
+    # owner: owner, repo: repo, branch: main, subdirectory: subdirectory
+
+    # gh:owner/repo@main/sub/directory
+    # owner: owner, repo: repo, branch: main, subdirectory: sub/directory
+
+    # gh:owner/repo/sub/directory
+    # owner: owner, repo: repo, branch: main, subdirectory: sub/directory
```
michal-stlv: This is lot of comments that should really be a docstring and explain why rather than what, e.g.:

```py
def _parse_template_string(template: str) -> tuple[str, str, str, str]:
    """Parse template string into GitHub repository components.

    Supports formats:
    - 'base' → stelviodev/templates (main branch)
    - 'gh:owner/repo' → owner/repo (main branch)
    - 'gh:owner/repo@branch' → with specific branch
    - 'gh:owner/repo/subdir' → with subdirectory
    - 'gh:owner/repo@branch/subdir' → branch + subdirectory

    Returns:
        Tuple of (owner, repo, branch, subdirectory)
      """
```

stelvio/cli/__init__.py
```diff
@@ -249,3 +250,57 @@ def determine_env(environment: str) -> str:
...
+        else:
+            repo_part = gh_template
+            branch = "main"
+            subdirectory = None
+
+        if "/" in repo_part:
+            parts = repo_part.split("/", 2)
+            owner = parts[0]
+            repo = parts[1]
+            if len(parts) == 3:  # noqa: PLR2004
```
michal-stlv: The noqa comment suppresses the magic number warning rather than addressing it.

Use a named constant or make it clearer. And consider to  add a descriptive comment:
```py
OWNER_REPO_SUBDIR_PARTS = 3 # owner/repo/subdirectory format`
if len(parts) == OWNER_REPO_SUBDIR_PARTS:
```

stelvio/cli/__init__.py
```diff
@@ -249,3 +250,57 @@ def determine_env(environment: str) -> str:
...
+            repo_part, branch_subdir_part = gh_template.split("@", 1)
+            if "/" in branch_subdir_part:
+                branch, subdirectory = branch_subdir_part.split("/", 1)
+            else:
+                branch = branch_subdir_part
+                subdirectory = None
+        else:
+            repo_part = gh_template
+            branch = "main"
+            subdirectory = None
```
michal-stlv, line 311: `subdirectory` is typed as str in the return type, but can be `None`. This should be str |
   None.

```py
def _parse_template_string(template: str) -> tuple[str, str, str, str | None]:
```

And we need to make sure to handle `None` properly in the calling code.

stelvio/cli/__init__.py
```diff
@@ -249,3 +250,57 @@ def determine_env(environment: str) -> str:
...
+        if "/" in repo_part:
+            parts = repo_part.split("/", 2)
+            owner = parts[0]
+            repo = parts[1]
+            if len(parts) == 3:  # noqa: PLR2004
+                subdirectory = parts[2]
+        else:
+            raise ValueError("Invalid template format")
+    else:
+        raise ValueError("Invalid template format")
```
michal-stlv: Can we have more specific error messages? For each specific case?

e.g.:
```py
else:
    raise ValueError(
        f"Invalid template format: '{repo_part}'. "
        "Expected format: gh:owner/repo[@branch][/subdirectory]"
    )
```

sebst:
> I guess no need to put anything in the docs until we have some useful templates?

Yes, not needed until we have templates

stelvio/cli/__init__.py
```diff
@@ -249,3 +250,57 @@ def determine_env(environment: str) -> str:
         user_env = getpass.getuser()
         save_user_env(user_env)
     return user_env
+
+
+OWNER_REPO_SUBDIR_PARTS = 3  # owner/repo/subdirectory format`
```
michal-stlv: there's a backtick at the end of this line which looks like typo? 
```py
OWNER_REPO_SUBDIR_PARTS = 3  # owner/repo/subdirectory format`<--- HERE

```

stelvio/cli/__init__.py
```diff
@@ -249,3 +250,57 @@ def determine_env(environment: str) -> str:
...
+        else:
+            raise ValueError(
+                f"Invalid template format: '{repo_part}'. "
+                "Expected format: gh:owner/repo[@branch][/subdirectory]"
+            )
+    else:
+        raise ValueError(
+            f"Invalid template format: '{repo_part}'. "
+            "Expected format: gh:owner/repo[@branch][/subdirectory]"
+        )
```
michal-stlv: Looking at this again, is this code reachable? I think it's not as first condition  is `if not template.startswith("gh:"):` and then `elif template.startswith("gh:"):` so I'd say `else` will never get executed? Or am I missing something?

michal-stlv (approved):

PR #61, opened 2025-10-26
feat: add `--version` flag to cli

sebst:
Closes #58

PR #66, opened 2025-10-27
feat: add `system` command to cli

### LINKED ISSUE #65: Add `system` command to CLI
`system` should execute basic sanity checks including installing Pulumi.

This is needed for unattended installations, e.g. in a Dockerfile

sebst:
Closes #65

### DIFF
--- a/stelvio/cli/__init__.py
+++ b/stelvio/cli/__init__.py
@@ -66,11 +66,11 @@ def safe_run_pulumi(func: Callable, env: str | None, **kwargs: bool) -> None:
 @click.option(
     "--verbose", "-v", count=True, help="Increase verbosity. -v for INFO, -vv for DEBUG logs."
 )
-@click.option("--version", is_flag=True, help="Show version and exit.")
+@click.option("--version", is_flag=True, help="Show Stelvio and Pulumi versions.")
 @click.pass_context
 def cli(ctx: click.Context, verbose: int, version: bool) -> None:
     if version:
-        _version()
+        _version()  # Exits after printing versions
 
     # If no command was invoked, show help
     if ctx.invoked_subcommand is None:
@@ -148,6 +148,8 @@ def system() -> None:
     """Performs a system check for Stelvio."""
     _ensure_pulumi()
     console.print("[green]✓[/green] System check passed")
+    """Shows Stelvio and Pulumi versions."""
+    _version()
 
 
 @click.command()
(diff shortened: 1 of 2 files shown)


stelvio/cli/__init__.py
```diff
@@ -131,15 +139,22 @@ def init(template: str | None) -> None:
...
-    """Prints the version of Stelvio."""
-    console.print(metadata.version("stelvio"))
+    """Shows version and exit."""
+    _version()
+
+
+@click.command()
+def system() -> None:
+    """Performs a system check for Stelvio."""
+    _ensure_pulumi()
```
michal-stlv, line 149: If Pulumi is already installed, this command does nothing and exits silently. User won't know if the check passed or failed? 

Maybe we could print something? 
at least:
```py
_ensure_pulumi()
console.print("[green]✓[/green] System check passed")
```

or more detailed but maybe it's too much. 
```py
 def system() -> None:
      """Performs a system check for Stelvio."""    
    console.print("[bold]Running system check...[/bold]\n")

    # Check Pulumi
    if needs_pulumi():
        console.print("[yellow]•[/yellow] Pulumi not found, installing...")
        install_pulumi()
        console.print("[green]✓[/green] Pulumi installed successfully")
    else:
        console.print("[green]✓[/green] Pulumi is already installed")

    console.print("\n[bold green]System check passed[/bold green]")
```

stelvio/cli/__init__.py
```diff
@@ -131,15 +139,22 @@ def init(template: str | None) -> None:
...
 def version() -> None:
-    """Prints the version of Stelvio."""
-    console.print(metadata.version("stelvio"))
+    """Shows version and exit."""
+    _version()
+
+
+@click.command()
+def system() -> None:
+    """Performs a system check for Stelvio."""
```
michal-stlv, line 148: We say we perform system check but we really check check pulumi? So maybe just say that in the docstring ? Or do we want to check more?

michal-stlv (changes requested):
I dont' see any `cli.add_command(system)` so this is not added? 
As with #61 would be good to document it in docs/guides/using-cli.md maybe mentioning its use in dockerfile etc?

stelvio/cli/__init__.py
sebst: We do wanna check more in the future, e.g. ensuring that we're in a proper venv

PR #70, opened 2025-10-29
feat: add `outputs` command to cli

### LINKED ISSUE #69: Add `output` command to CLI

sebst:
Closes #69

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,1 @@
+FROM mcr.microsoft.com/devcontainers/python:3.13-trixie
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,14 +1,29 @@
-// For format details, see https://aka.ms/devcontainer.json. For config options, see the
-// README at: https://github.com/devcontainers/templates/tree/main/src/python
 {
 	"name": "Stelvio.dev (py3.13, trixie)",
-	"image": "mcr.microsoft.com/devcontainers/python:3.13-trixie",
+	"build": {
+		"dockerfile": "./Dockerfile",
+		"context": "..",
+		"cacheFrom": "type=registry,ref=ghcr.io/stelviodev/stelviodevcontainer:latest"
+	},
 	"features": {
 		"ghcr.io/devcontainers/features/aws-cli:1": {},
 		"ghcr.io/devcontainer-community/devcontainer-features/bun.sh:1": {},
-		// "ghcr.io/devcontainers-extra/features/pulumi:1": {}, Pulumi is not needed anymore since stlv installs its own
-		"ghcr.io/devcontainers/features/github-cli:1": {},
-		"ghcr.io/va-h/devcontainers-features/uv:1": {}
+		"ghcr.io/devcontainer-community/devcontainer-features/uv:1": { "shellautocompletion": "true" },
+		"ghcr.io/devcontainers/features/github-cli:1": {}
+	},
+	"customizations": {
+		"vscode": {
+			"extensions": [
+				"ms-python.python"
+			],
+			"settings": {
+				"python.testing.pytestArgs": [
+					"tests"
+				],
+				"python.testing.unittestEnabled": false,
+				"python.testing.pytestEnabled": true
+			}
+		}
 	},
 	"postCreateCommand": ".devcontainer/setup",
 	"containerEnv": {
--- a/.devcontainer/setup
+++ b/.devcontainer/setup
@@ -32,6 +32,6 @@ UV_VENV_CLEAR=1 uv venv
 # (Removed redundant uv pip install --editable .)
 
 # Run a sync
-uv sync
+uv sync --dev
 
 echo " (*) Done."
--- a/.github/workflows/prebuild-devcontainer.yml
+++ b/.github/workflows/prebuild-devcontainer.yml
@@ -1,0 +1,43 @@
+name: Pre-build Devcontainer
+
+on:
+    workflow_dispatch:
+    push:
+        branches:
+        - main
+
+jobs:
+    build_devcontainer:
+        runs-on: ubuntu-24.04
+        permissions:
+            contents: read
+            packages: write
+        steps:
+        - name: Checkout code
+          uses: actions/checkout@v4
+
+        - name: Set up QEMU
+          uses: docker/setup-qemu-action@v3
+
+        - name: Set up Docker Buildx
+          uses: docker/setup-buildx-action@v3
+          with:
+            version: v0.29.1
+            platforms: linux/amd64,linux/arm64
+            install: true
+            use: true
+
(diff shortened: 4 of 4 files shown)


stelvio/pulumi.py
```diff
@@ -298,6 +299,47 @@ def run_pulumi_refresh(environment: str | None) -> None:
         raise SystemExit(1) from None
 
 
+def run_pulumi_output(environment: str | None, json: bool = False) -> None:
```
michal-stlv: `environment` can't be none can it? we don't allow or expect that so type annotation shoudl reflect that and be just `environment: str`

docs/guides/using-cli.md
```diff
@@ -169,6 +170,13 @@ stlv destroy           # remove your personal environment
 stlv destroy staging   # remove staging environment
 ```
 
+### Showing Pulumi outputs
+
+```bash
+stlv output            # Shows a list of Pulumi outputs
+stlv output --json     # Shows a list of Pulumi outputs in JSON format
```
michal-stlv: I'm not sure "a list of" is technically correct?  it's key value pairs but mainly we should mention it's outputs of selected environment  so maybe:
```
stlv output            # Shows environment outputs (key-value pairs)
stlv output --json     # Shows environment outputs in JSON format
```

stelvio/cli/__init__.py
```diff
@@ -227,17 +240,31 @@ def unlock(env: str | None) -> None:
...
     env = determine_env(env)
     safe_run_pulumi(run_pulumi_cancel, env)
 
 
+@click.command()
+@click.argument("env", default=None, required=False)
+@click.option("--json", is_flag=True, help="Output in JSON format")
+def output(env: str | None, json: bool) -> None:
+    """
+    Outputs the current state of your app.
```
michal-stlv: We can do better here. It doesn't ouputs state. it shows pulumi outputs for selected environment. outputs are not state. Maybe something like this?
```
Shows environment outputs in key-value pairs or in JSON format if `json` param is `True`
```

stelvio/pulumi.py
```diff
@@ -298,6 +299,47 @@ def run_pulumi_refresh(environment: str | None) -> None:
...
+        with console.status("Loading app..."):
+            load_stlv_app()
+            app = StelvioApp.get_instance()
+            app_name = app._name  # noqa: SLF001
+            stack = prepare_pulumi_stack(environment)
+
+        try:
+            outputs = stack.outputs()
+            if outputs:
+                console.print_json(data={key: value.value for key, value in outputs.items()})
```
michal-stlv, line 314: Don't we want to include in the json environment these outputs are for?  So user/system knows the source? Not saying it's a must just thinking loud if that might be useful or necessary in some cases. Or we assume that since caller knows env then it's their responsibility...

stelvio/pulumi.py
```diff
@@ -298,6 +299,47 @@ def run_pulumi_refresh(environment: str | None) -> None:
         raise SystemExit(1) from None
 
 
+def run_pulumi_output(environment: str | None, json: bool = False) -> None:
+    # For JSON output, skip the handler to avoid spinner and header output
+    if json:
+        with console.status("Loading app..."):
```
michal-stlv, line 305: We skip spinenr and header etc. but we still pritn "Loading app..." doesn't it interfere with json output? just check if it's intentional.

stelvio/cli/__init__.py
```diff
@@ -253,13 +254,26 @@ def unlock(env: str | None) -> None:
     safe_run_pulumi(run_pulumi_cancel, env)
 
 
+@click.command()
+@click.argument("env", default=None, required=False)
+@click.option("--json", is_flag=True, help="Output in JSON format")
+def output(env: str | None, json: bool) -> None:
```
michal-stlv: I think command should be called `outputs` - notice `s` at the end. This makes sense to me more as it prints outputS. it's not about outputting anything, it's specifically about printing outputs (of the last deploy). Also pulumi's method is outputs in the same line of thinking. can you rename please or you had some specific arguments for naming it output (without s)?

stelvio/pulumi.py
sebst: Most likely the json output is going to be used in some shell script, maybe piped through jq or something like that. So the entire output should be valid JSON without additions

stelvio/pulumi.py
sebst: Had the same question -- I tested it with jq, and it works. The spinner gets cleared, so I left it to have some interactivity

stelvio/cli/__init__.py
sebst: You could argue that the output is one output object, and I kept semantics aligned with [pulumi's](https://www.pulumi.com/docs/iac/cli/commands/pulumi_stack_output/).

I prefer `output` for that reason, but I'm fine with `outputs` as well. I'll let you decide

stelvio/pulumi.py
sebst: It can't be? The other `run_pulumi_*` functions have `| None` as well....

stelvio/cli/__init__.py
michal-stlv: interesting. they have `pulumi stack output` in cli but `stack.outputs()` in Automation API. If you're fine with both I'd prefer `outputs`. Just sounds right to me.

stelvio/pulumi.py
michal-stlv: hmmm i think during some changes it was changed and now we always pass value.  before i think default env was inside these functions but now it's all determined outside. so not really scope of this PR if you dont' want to do it.

sebst: Added issue: #73

stelvio/cli/__init__.py
sebst: Ok, renamed it

sebst:
Need manual testing in codespaces, but they're down rn 🤣

stelvio/pulumi.py
```diff
@@ -298,6 +299,47 @@ def run_pulumi_refresh(environment: str | None) -> None:
...
+            outputs = stack.outputs()
+            if outputs:
+                console.print_json(data={key: value.value for key, value in outputs.items()})
+        except CommandError as e:
+            console.print(f"[red]{e!s}[/red]")
+            if os.getenv("STLV_DEBUG", "0") == "1":
+                raise e  # noqa: TRY201
+            raise SystemExit(1) from None
+    else:
+        stack, app_name, handler = setup_operation(environment, "output")
```
michal-stlv: last param needs to be `outputs` with S at the end.

stelvio/pulumi.py
```diff
@@ -182,6 +182,7 @@ def setup_operation(
         "destroy": "Destroying",
         "refresh": "Refreshing",
         "unlock": "Unlocking",
+        "output": "Outputs for",
```
michal-stlv: and here also `outputs`?

PR #72, opened 2025-10-29
Improve user experience in devcontainer, prebuild…

### LINKED ISSUE #68: Improve devcontainer
- cacheFrom / cacheTo
- VSCode settings for pytest
- VSCode settings for extensions
- sync --dev / sync 
- separate devcontainer for stelvio dev and stelvio users
- prebuild in ci

sebst:
Closes #68


sebst:
Draft!!

.devcontainer/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM mcr.microsoft.com/devcontainers/python:3.13-trixie
```
michal-stlv: missing newline

.github/workflows/prebuild-devcontainer.yml
```diff
@@ -0,0 +1,41 @@
...
+          with:
+            registry: ghcr.io
+            username: ${{ github.repository_owner }}
+            password: ${{ secrets.GITHUB_TOKEN }}
+
+        - name: Pre-build dev container image
+          uses: devcontainers/ci@v0.3
+          with:
+            imageName: ghcr.io/stelviodev/stelviodevcontainer
+            # cacheTo: type=inline  # This is the default value.
```
michal-stlv, line 38: This is commented out but issue #68 mentions "cacheFrom / cacheTo". The devcontainer uses cacheFrom but the build workflow doesn't explicitly set cacheTo.

Is the default type=inline sufficient, or should this be explicit? If default is fine, remove the comment? If not, uncomment and configure properly.

.github/workflows/prebuild-devcontainer.yml
```diff
@@ -0,0 +1,41 @@
...
+            runCmd: sleep 1
```
michal-stlv, line 39: The runCmd just sleeps. There's no verification that the container actually works correctly after building.

We could run a basic sanity check?:
```  
runCmd: |
    python --version && \
    uv --version && \
    aws --version && \
    gh --version
```
  This ensures all features are properly installed.

.github/workflows/prebuild-devcontainer.yml
```diff
@@ -0,0 +1,41 @@
...
+        - name: Checkout code
+          uses: actions/checkout@v4
+
+        - name: Set up QEMU
+          uses: docker/setup-qemu-action@v3
+
+        - name: Set up Docker Buildx
+          uses: docker/setup-buildx-action@v3
+          with:
+            version: v0.16.1
```
michal-stlv: Why pin to this specific version? Is there a reason not to use latest or at least a more recent version? (Current latest is v0.17+)

.github/workflows/prebuild-devcontainer.yml
```diff
@@ -0,0 +1,41 @@
...
+            platforms: linux/amd64,linux/arm64
```
michal-stlv, line 23: is this necessay here? actual platform list  for container are set on line 41 no?

.devcontainer/devcontainer.json
```diff
@@ -1,14 +1,29 @@
-// For format details, see https://aka.ms/devcontainer.json. For config options, see the
-// README at: https://github.com/devcontainers/templates/tree/main/src/python
 {
 	"name": "Stelvio.dev (py3.13, trixie)",
-	"image": "mcr.microsoft.com/devcontainers/python:3.13-trixie",
+	"build": {
+        "dockerfile": "./Dockerfile",
+        "context": "..",
+        "cacheFrom": "type=registry,ref=ghcr.io/stelviodev/stelviodevcontainer:latest"
+    },
```
michal-stlv: Here you use spaces but rest of file uses tabs for indentation? should be all same...

michal-stlv (commented):
Issue #68 mentions: separate devcontainer for stelvio dev and stelvio users
This isn't addressed in the PR. Is this intentional (for a future PR)?

Would be good to have some docs about:
- How to use the pre-built container
- What VS Code settings are pre-configured
- Why users might want --dev flag

Maybe add .devcontainer/README.md or update project docs or ideally both?

.github/workflows/prebuild-devcontainer.yml
sebst: I prefer it to be explicit, but adding it explicitly causes a weird bug, hence I left the comment

.github/workflows/prebuild-devcontainer.yml
sebst: This imo is not the right place for sanity checks of the features. They fail during installation if something goes wrong. I intentionally used features from the "most official" repos, and these features come with the sanity checks in their own pipeline.

michal-stlv: hmmm ok. i meant it'd check that container is build properly but sure i get it.

.github/workflows/prebuild-devcontainer.yml
sebst: Current latest is actually 0.29 . Changed it

.github/workflows/prebuild-devcontainer.yml
sebst: This is the setup action, which relies on qemu. [Docs](https://github.com/docker/setup-buildx-action?tab=readme-ov-file#inputs) say: "If not empty, values take priority over the detected ones" -- actually not sure if "detected ones" would add the capability to run diff platforms.

I'd suggest keeping it, since it is [tested](https://devcontainer.community/20250303-prebuild-devcontainer/)

PR #76, opened 2025-10-30
Make `environment` argument non-optional in `run_pulumi_*` functions

### LINKED ISSUE #73: Make `environment` argument non-optional in `run_pulumi_*` functions
Remove `| None` type annotation from  `run_pulumi_*` functions in `pulumi.py`

sebst:
Closes #73

### DIFF
--- a/stelvio/cli/__init__.py
+++ b/stelvio/cli/__init__.py
@@ -278,7 +278,7 @@ cli.add_command(outputs)
 cli.add_command(system)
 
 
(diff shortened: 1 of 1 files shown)


stelvio/cli/__init__.py
michal-stlv: not related to the issue but can we do `environment: str | None` here? My mistake from before.

sebst: fixed

Issue #48, opened 2025-10-16
Router: Support S3 + API Gateway on Same Custom Domain


michal-stlv:
## Problem

Common SPA pattern: serve static files from S3 and API from API Gateway on the same domain to avoid CORS.

**Desired:**
```
https://app.example.com/           → S3 (index.html)
https://app.example.com/about      → S3 (SPA routing)
https://app.example.com/api/users  → API Gateway
```

**Currently not possible** because:
- `CloudFrontDistribution` only supports single S3 origin
- API Gateway custom domains are direct (not through CloudFront)
- Can't combine both on one domain

## Why This Matters

**With same domain (desired):**
- ✅ No CORS configuration needed
- ✅ Single SSL certificate
- ✅ Single custom domain
- ✅ Better performance (CloudFront CDN)

**With separate domains (current workaround):**
- ❌ Requires CORS setup
- ❌ Two certificates, two DNS records
- ❌ Preflight overhead
- ❌ Cookie/auth complications

## High-Level Solution

Enhance CloudFrontDistributon to support:
1. Multiple origins (S3 + API Gateway)
2. Path-based routing behaviors (`/api/*` → API Gateway, `/*` → S3)
3. Different caching policies per path

Not sure how exactly it should work from user perspective yet. But Some kind of component that would bind bucket and api under one domain. 

## Implementation Notes

- API Gateway should use REGIONAL endpoint (not edge-optimized)
- Disable caching for `/api/*` paths
- SPA client-side routing needs CloudFront Function (not custom error pages, which would affect API 404s)
- API design TBD

## References

- [AWS: SPA + API Gateway Same Domain](https://aws.amazon.com/blogs/networking-and-content-delivery/improve-single-page-application-spa-performance-with-a-same-domain-policy-using-amazon-cloudfront/)

michal-stlv:
Hey @sebst . One more thing. I've been thinking about CloudFront caching and wanted to raise it for discussion.

Current defaults hardcode TTLs per origin type:
- **S3 buckets**: `default_ttl: 300` (5 min), `max_ttl: 3600` (1 hour)
- **API Gateway/Lambda**: `default_ttl: 0` (no cache)

API Gateway/Lambda with TTL=0 makes sense - dynamic responses shouldn't cache.

[...]

PR #88, opened 2025-11-21
feat: Function URLs for Lambdas in `Router`

sebst:


sebst:
Changes for https://github.com/stelviodev/stelvio/issues/48#issuecomment-3493628070

stelvio/aws/cloudfront/origins/lambda_function.py
```diff
@@ -15,22 +15,80 @@ def __init__(self, idx: int, route: Route) -> None:
...
+            return normalized
+        raise TypeError(f"Invalid function_url config type: {type(config).__name__}")
+
     def get_origin_config(self) -> RouterRouteOriginConfig:
+        # Normalize function URL configuration
+        url_config = self._normalize_function_url_config(self.route.function_url_config)
+
+        # Determine authorization type
+        # auth='iam' → 'AWS_IAM', auth=None → 'NONE'
+        auth_type = "AWS_IAM" if url_config.auth == "iam" else "NONE"
```
michal-stlv, line 49: Wouldn't it be simpler if, in `_normalize_function_url_config`, we would not deal with auth at all? Just went with whatever is there, so you would end up without default none or IAM. And then in `get_origin_config` on line 49 you would check if auth is none, and if it is none, then you would use none; in all other cases, you would just use aws_iam.

stelvio/aws/cloudfront/origins/lambda_function.py
```diff
@@ -15,22 +15,80 @@ def __init__(self, idx: int, route: Route) -> None:
...
+                "allow_credentials": normalized_cors.allow_credentials,
+                "allow_headers": to_list(normalized_cors.allow_headers),
+                "allow_methods": to_list(normalized_cors.allow_methods),
+                "allow_origins": to_list(normalized_cors.allow_origins),
+                "expose_headers": normalized_cors.expose_headers or [],
+                "max_age": normalized_cors.max_age,
+            }
+
+        # Determine invoke mode based on streaming
+        invoke_mode = "RESPONSE_STREAM" if url_config.streaming else "BUFFERED"
```
michal-stlv, line 69: This is basically copy paste from `_create_function_url` in function.py I'm wondering if maybe this could be extracted and used in both places. Here we import stuff from the function, right? So we might as well just import a helper function to do this. Looking at it, actually, maybe the whole `_create_function_url` method could be turned into a function which just takes some parameters and does everything. In the existing version, it is a method which takes self but it uses only name from the self, so that could be param. And maybe there could be also a parameter for the auth.

stelvio/aws/cloudfront/origins/lambda_function.py
```diff
@@ -93,12 +155,32 @@ def get_origin_config(self) -> RouterRouteOriginConfig:
...
+
+        # Grant cloudfront.amazonaws.com permission to invoke via Function URL
+        return pulumi_aws.lambda_.Permission(
+            context().prefix(f"{self.function.name}-cloudfront-permission-{self.idx}"),
+            action="lambda:InvokeFunctionUrl",
+            function=self.function.resources.function.name,
+            principal="cloudfront.amazonaws.com",
+            source_arn=distribution.arn,
+            opts=pulumi.ResourceOptions(
+                depends_on=[distribution, self.function.resources.function]
```
michal-stlv: Do we need to explicitly say we depend on distribution? I guess not because we already passed distribution.arn as a source arn parameter, so that's an implicit dependency. Which will be handled by Pulumi. And I guess same is true for self.function.resources.function, which you also pass here as a parameter on line line with 18.

stelvio/aws/cloudfront/dtos.py
```diff
@@ -1,10 +1,13 @@
 from dataclasses import dataclass
-from typing import final
+from typing import TYPE_CHECKING, final
 
 import pulumi_aws
 
 from stelvio.component import Component
 
+if TYPE_CHECKING:
+    from stelvio.aws.function import FunctionUrlConfig, FunctionUrlConfigDict
```
michal-stlv: What is the point of this `if TYPE_CHECKING` I don't get it. It's okay to import in any case, not only when we are type checking, right?

stelvio/aws/cloudfront/router.py
```diff
@@ -17,6 +18,7 @@
 
 if TYPE_CHECKING:
     from stelvio.aws.cloudfront.cloudfront import CloudfrontPriceClass
+    from stelvio.aws.function import FunctionUrlConfig, FunctionUrlConfigDict
```
michal-stlv, line 21: Same here, if type_checking. I'm not getting why we need it here.

stelvio/aws/cloudfront/dtos.py
sebst: That's actually `TC001: Move application import `stelvio.aws.function.FunctionUrlConfigDict` into a type-checking block`

stelvio/aws/cloudfront/router.py
sebst: Same  `TC001`

PR #89, opened 2025-11-21
fix: respecting system aws profile and region and better this documen…

michal-stlv:
…ting behavior

### DIFF
--- a/docs/guides/troubleshooting.md
+++ b/docs/guides/troubleshooting.md
@@ -122,7 +122,7 @@ You have three options:
    @app.config
    def configuration(env: str) -> StelvioAppConfig:
     return StelvioAppConfig(
-        aws=AwsConfig(profile="michal"),
+        aws=AwsConfig(profile="michael"),
     )
    ```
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -76,7 +76,7 @@ select = ["E", "F", "I", "B", "N", "UP", "C4", "A", "S", "ARG", "LOG", "G", "PIE
     "PGH", "PL", "FURB", "RUF", "TRY", "INP", "ANN"
     # "D" - will need, defo need to work on more docs
 ]
-ignore = ['TRY003']
+ignore = ['TRY003', 'TRY300']
```
sebst: Why would [try-consider-else (TRY300)](https://docs.astral.sh/ruff/rules/try-consider-else/#try-consider-else-try300) be ignored?

docs/guides/troubleshooting.md
```diff
@@ -122,10 +122,7 @@ You have three options:
    @app.config
    def configuration(env: str) -> StelvioAppConfig:
     return StelvioAppConfig(
-        aws=AwsConfig(
-            region="us-east-1",
-            profile="michal",
-        ),
+        aws=AwsConfig(profile="michal"),
```
sebst: np: consider a more generic example name, e.g. `acme`, `alice`, or `bob`

michal-stlv: i don't like generic... but can do michael so it's proper english :)

pyproject.toml
michal-stlv: hmmm ok let's to those try/except/else...

PR #95, opened 2025-12-08
feat: s3 state sync

michal-stlv:


stelvio/aws/home.py
```diff
@@ -0,0 +1,104 @@
+import hashlib
+from pathlib import Path
+
+import boto3
+from botocore.exceptions import ClientError
+
+
+class AwsHome:
```
sebst, line 8: Wouldn't that be part of `stelvio.aws`?

michal-stlv: I moved it

docs/guides/environments.md
```diff
@@ -34,57 +31,30 @@ conflicts.
...
-stlv deploy prod
-```
-
-### Preview Changes by Environment
-
-```bash
-# See what would change in staging
-stlv diff staging
+!!! info
+    The `.stelvio/` folder contains personal settings and caches - add it to `.gitignore`.
```
sebst, line 45: Unrelated to PR but`stlv init` should ask to init git repo and create `.gitignore`

mkdocs.yml
```diff
@@ -62,7 +62,7 @@ nav:
   - Getting Started:
       - Quick Start: getting-started/quickstart.md
   - Guides:
-      - StelvioApp and stlv_app.py: guides/stelvio-app.md
```
sebst, line 65: Can this be in backticks?

docs/guides/environments.md
michal-stlv: feel free to create issue for that. although i'm not sure we should do that? does any other tools e.g. pulumi itself do this kind of stuff?

mkdocs.yml
michal-stlv: this is removed line, now it's `- Project Configuration: guides/stelvio-app.md` so does not apply

docs/guides/environments.md
sebst: Some tools do, some do not. `wrangler` does it, for example.

f'up to #96

PR #110, opened 2025-12-28
feat: Add support for AWS SESv2

### LINKED ISSUE #109: Add SES support
In `stlv_app.py`:
```python
@app.run
def run() -> None:
    email = Email(
        "stlvEmail",
        "sender@example.com",
        dmarc=None,
    )

    linked_function = Function(
        "MyFunctionA",
        handler="functions/api.handler",
        url="public",
        links=[email],
    )
```

In `api.py`:
```python
SENDER = StlvemailResource().email_identity_sender
client = boto3.client('sesv2')

...

response = client.send_email(
    FromEmailAddress=SENDER,
    Destination={
        'ToAddresses': [RECIPIENT]
    },
    Content={
        'Raw': {
            'Data': body
        }
    }
)
```

sebst:
closes #109

### DIFF
--- a/stelvio/aws/email.py
+++ b/stelvio/aws/email.py
@@ -65,7 +65,7 @@ class EmailConfig:
     """Typed configuration for the Email component."""
 
     sender: str
-    dmarc: str | None
+    dmarc: str | None = None
     events: list[EventConfiguration] | None = None
     sandbox: bool = False
     dns: Dns | Literal[False] | None = None


stelvio/aws/email.py
```diff
@@ -0,0 +1,286 @@
...
+    "subscription",
+    "open",
+    "click",
+]
+
+
+@final
+@dataclass(frozen=True)
+class EmailResources:
+    sender: str
```
michal-stlv: not pulumi resource should not be here

stelvio/aws/email.py
```diff
@@ -0,0 +1,286 @@
...
+    identity: pulumi_aws.sesv2.EmailIdentity
+    sandbox: bool = False
```
michal-stlv: not pulumi resource should not be here

stelvio/aws/email.py
```diff
@@ -0,0 +1,286 @@
...
+    """Dictionary configuration for the Email component."""
+
+    sender: str
+    dmarc: str | None | Literal[False]
+    events: list[EventConfiguration] | None = None
+    sandbox: bool = False
+    dns: Dns | Literal[False] | None = None
+
+
+@dataclass(frozen=True)
```
michal-stlv: we miss `kw_only=True` which we have for all config dataclasses

stelvio/aws/email.py
```diff
@@ -0,0 +1,286 @@
...
+    event_destinations: list[pulumi_aws.sesv2.ConfigurationSetEventDestination] | None = None
+
+
+class EventConfiguration(TypedDict):
+    name: str
+    types: list[EventType]
+    topic_arn: str
+
+
+class EmailConfigDict(TypedDict):
```
michal-stlv: missing `total=False` which we have for all *ConfigDict classes

stelvio/aws/email.py
```diff
@@ -0,0 +1,286 @@
...
+        elif config.dmarc is False:
+            config = EmailConfig(
+                sender=config.sender,
+                dmarc=None,
+                events=config.events,
+                sandbox=config.sandbox,
+                dns=config.dns,
+            )
+
+        return config
```
michal-stlv, line 168: The pattern differs from other components e.g. DynamoTable._parse_config. DynamoTable raises error when both config and opts are provided:
```py
  # DynamoTable pattern:
  if config and opts:
      raise ValueError("cannot combine 'config' parameter with additional options")
```
  But Email allows mixing them, which is inconsistent with the codebase and also confusing as what are we applying than?

stelvio/aws/email.py
```diff
@@ -0,0 +1,286 @@
...
+        link_creator_ = ComponentRegistry.get_link_config_creator(type(self))
+
+        link_config = link_creator_(self)
+        return Link(self.name, link_config.properties, link_config.permissions)
+
+
+@link_config_creator(Email)
+def default_email_link(
+    email: Email,
+) -> LinkConfig:
```
michal-stlv, line 267: this should receive pulumi resource not stelvio component to be consistent also this breaks type checking. 
```py
def link_config_creator[T: PulumiResource](
    component_type: type[Component],
) -> Callable[[Callable[[T], LinkConfig]], Callable[[T], LinkConfig]]:
    """Decorator to register a default link creator for a component type"""

    def decorator(func: Callable[[T], LinkConfig]) -> Callable[[T], LinkConfig]:
        @wraps(func)
        def wrapper(resource: T) -> LinkConfig:
            return func(resource)

        ComponentRegistry.register_default_link_creator(component_type, func)
        return wrapper

    return decorator
```
i see you need more pulumi resources to create link .. so maybe we need to update link creator types etc.? or find other way.

stelvio/aws/email.py
```diff
@@ -0,0 +1,286 @@
...
+            "email_identity_sender": identity.email_identity,
+            "email_identity_arn": identity.arn,
+            "configuration_set_name": configuration_set.configuration_set_name,
+            "configuration_set_arn": configuration_set.arn,
+        },
+        permissions=[
+            AwsPermission(
+                actions=["ses:*"],
+                resources=[identity.arn, configuration_set.arn],
+            ),
```
michal-stlv: This grants ALL SES permissions including administrative actions like DeleteEmailIdentity, DeleteConfigurationSet, etc. We need to narrow it down  to what's actually needed. it also makes permission below irellevant as this is broader.

docs/guides/email.md
```diff
@@ -0,0 +1,166 @@
...
+from stelvio.aws.email import Email
+from stelvio.aws.function import Function
+
+@app.run
+def run() -> None:
+    # Create an email identity
+    email = Email(
+        "stlvEmail",
+        "sender@example.com",
+        dmarc=None,
```
michal-stlv: Example shows dmarc=None but doesn't explain this means "use default DMARC policy" for domains. Confusing since None usually means "no value". If it's default maybe just don't set it?

docs/guides/email.md
```diff
@@ -0,0 +1,166 @@
...
+import boto3
+from stlv_resources import StlvemailResource
+
+def handler(event, context):
+    client = boto3.client('sesv2')
+    
+    # Access the linked resource properties
+    # The class name is derived from the component name (stlvEmail -> StlvemailResource)
+    resources = StlvemailResource()
+    SENDER = resources.email_identity_sender
```
michal-stlv: what's this? that's not how. resources work. 

you do `from stlv_resources import Resources`

then e.g. `Resources.todos` if my resource is named `todos`

tests/aws/email/test_email.py
```diff
@@ -0,0 +1,404 @@
...
+        # First permission: ses:* on identity and config set
+        perm1 = perms[0]
+        assert perm1.actions == ["ses:*"]
+        assert len(perm1.resources) == 2
+
+        # Second permission: sending emails
+        perm2 = perms[1]
+        expected_actions = ["ses:SendEmail", "ses:SendRawEmail", "ses:SendTemplatedEmail"]
+        assert sorted(perm2.actions) == sorted(expected_actions)
+        assert len(perm2.resources) == 1
```
michal-stlv: are we precise with assertions?  we shoudl check for specific values e.g. exact arn etc. not just counts

stelvio/aws/email.py
```diff
@@ -0,0 +1,286 @@
...
+            )
+        if self.is_domain:
+            self.check_domain(self.config.sender)
+        else:
+            self.check_email(self.config.sender)
+
+        self._resources = None
+
+    @property
+    def config(self) -> EmailConfig:
```
michal-stlv, line 112: we have this config property here, in api and in function but not in other components. just noting in consistency mainly to myself.  also thinking if it's even useful...

tests/aws/email/test_email.py
```diff
@@ -0,0 +1,404 @@
...
+
+        if args.typ == "aws:sesv2/emailIdentity:EmailIdentity":
+            props["dkim_signing_attributes"] = {"tokens": ["token1", "token2", "token3"]}
+            props["arn"] = (
+                f"arn:aws:ses:us-east-1:123456789012:identity/{args.inputs['emailIdentity']}"
+            )
+
+        if args.typ == "aws:sesv2/configurationSet:ConfigurationSet":
+            props["arn"] = "arn:aws:ses:us-east-1:123456789012:configuration-set/"
+            f"{args.inputs['configurationSetName']}"
```
michal-stlv: This will not join the strings will it? Isn't these two separate statements ?   we need \ or parenthesis no?

michal-stlv (changes requested):
We're missing some important tests 
- Sandbox mode link permissions - IAM difference untested
- Event destinations -  zero tests?
- DMARC=False explicitly disabling DMARC
- Configuration set properties in link verification

stelvio/aws/email.py
```diff
@@ -0,0 +1,286 @@
...
+
+@final
+@dataclass(frozen=True)
+class EmailResources:
+    sender: str
+    identity: pulumi_aws.sesv2.EmailIdentity
+    sandbox: bool = False
+    configuration_set: pulumi_aws.sesv2.ConfigurationSet | None = None
+    dkim_records: list[pulumi_aws.route53.Record] | None = None
+    dmarc_record: pulumi_aws.route53.Record | None = None
```
michal-stlv: type annotations are route53.Record] but self.dns.create_record can return also cloudflare... do we assume only route53 used here or possible to make it work with cloudflare?

sebst: Fixed

docs/guides/email.md
sebst: fixed

docs/guides/email.md
sebst: fixed

stelvio/aws/email.py
sebst: disabled for now. could you test with a sandboxed account? that was the initial reason.... mine is not in sandbox anymore

stelvio/aws/email.py
sebst: I thought we'd refactor that in the to use the `*Resources` class?

stelvio/aws/email.py
sebst: fixed

stelvio/aws/email.py
sebst: fixed

stelvio/aws/email.py
sebst: kept for linking

stelvio/aws/email.py
sebst: fixed

stelvio/aws/email.py
sebst: fixed

tests/aws/email/test_email.py
sebst: fixed

docs/guides/email.md
```diff
@@ -0,0 +1,163 @@
...
+
+```python
+from stelvio.aws.email import Email
+from stelvio.aws.function import Function
+
+@app.run
+def run() -> None:
+    # Create an email identity
+    email = Email(
+        "stlvEmail",
```
michal-stlv: nitpick but I'd assume pythonistas don't use camelCase to name things? rather kebab-case is common in aws or maybe they'd use snake_case

docs/guides/email.md
```diff
@@ -0,0 +1,163 @@
...
+
+```python
+import boto3
+from stlv_resources import Resources
+
+def handler(event, context):
+    client = boto3.client('sesv2')
+    
+    # Access the linked resource properties
+    resources = Resources.stlvEmail
```
michal-stlv: stelvio generates snake_case properties so stlv_email should be resulting property name

tests/aws/email/test_email.py
michal-stlv: this still needs fixing

stelvio/aws/email.py
michal-stlv: you pass whole email component to the link creator so you don't need it in resources.

stelvio/aws/email.py
michal-stlv: yes but would be good to do in this PR so we're not in incorrect state. now Email component is passing itself, to link config creator, but bucket and dynamo table are passing single pulumi resources. thinking about this passign *Resources is not better than passing component itself. So I suggest we update this decorator signature and existing links creators functions (just bucket and dynamo table)  and link methods to work like Email in this PR and we're done with it. 
Happy to do it if you don't feel like it.

stelvio/aws/email.py
```diff
@@ -0,0 +1,283 @@
...
+                dns=config.dns,
+            )
+        # Then handle explicit dmarc=False to disable DMARC
+        elif config.dmarc is False:
+            config = EmailConfig(
+                sender=config.sender,
+                dmarc=None,
+                events=config.events,
+                dns=config.dns,
+            )
```
michal-stlv, line 166: If user does `Email("x", "example.com", dmarc=False, sandbox=True, dns=...)`, the `sandbox=True` gets lost (defaults to `False`). Is this intentional ?

stelvio/aws/email.py
michal-stlv: i tested and i was able to send and receive email

stelvio/link.py
```diff
@@ -97,3 +97,13 @@ def remove_properties(self, *keys: str) -> "Link":
 class Linkable(Protocol):
     def link(self) -> Link:
         raise NotImplementedError
+
+
+class LinkableMixin:
+    def link(self) -> Link:
+        from stelvio.component import ComponentRegistry
```
michal-stlv, line 104: i guess this is necessary due to circular imports and there's  no better way to do it?

sebst: exactly

PR #113, opened 2025-12-28
fix: handle imports in dev mode correctly

### LINKED ISSUE #112: `stlv dev` does not pick up `stlv_resources.py`
When running `stlv dev` with a function that has a `stlv_resources.py`, this causes an `ImportError` as the file is not picked up by `runpy`.

sebst:
closes #112

### DIFF
--- a/stelvio/aws/function/function.py
+++ b/stelvio/aws/function/function.py
@@ -265,7 +265,6 @@ class Function(Component[FunctionResources], BridgeableComponent):
         handler_file_path = project_root / handler_file
         handler_function_name = self.config.handler_function_name
 
-        sys.path.insert(0, str(handler_file_path.parent))
         new_environ = await self._get_environment_for_bridge_event()
 
         with temporary_environment(new_environ, [handler_file_path.parent]):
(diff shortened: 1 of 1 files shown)


stelvio/aws/function/function.py
```diff
@@ -261,15 +263,21 @@ async def _handle_bridge_event(self, data: dict) -> BridgeInvocationResult | Non
...
-            module = runpy.run_path(str(handler_file_path))
-        except FileNotFoundError:
-            logger.exception(
-                "Function handler file not found: %s (expected at %s)",
-                handler_file,
-                handler_file_path,
-            )
-            return None
+            try:
+                module = runpy.run_path(str(handler_file_path))
```
michal-stlv, line 272: The environment variables are set AFTER `runpy.run_path()` executes.

While stlv_resources.py uses @cached_property (so env vars are read on first access, not import), users who access Resources.xxx.yyy at module level will get `None`:

```py
# This pattern will FAIL in dev mode with this PR:
from stlv_resources import Resources
TABLE_NAME = Resources.my_table.name  # Read at import time - env vars not set yet!

def handler(event, context):
    ...
```

Environment variables must be set BEFORE calling `runpy.run_path()`.

stelvio/aws/function/function.py
```diff
@@ -278,16 +286,28 @@ async def _handle_bridge_event(self, data: dict) -> BridgeInvocationResult | Non
...
+                        None, function, event.get("event", {}), lambda_context
+                    )
+                except Exception as e:
+                    error = e
+                end_time = time.perf_counter()
+            finally:
+                # Restore environment
+                os.environ.clear()
+                os.environ.update(original_environ)
+                new_environ.clear()
```
michal-stlv: this is pointless no? why clean ? it's local variable which will be garbage collected.

michal-stlv (approved):

PR #114, opened 2025-12-29
feat: Add `Queue` (SQS)

sebst:


michal-stlv (commented):
This misses the main thing - handler. Function ( and later container etc.) which will handles messages from queue. Also DL queue setting etc

docs/guides/queues.md
```diff
@@ -0,0 +1,319 @@
...
+    - Successfully processed messages are automatically deleted
+    - Failed messages eventually move to the DLQ (if configured)
+
+## Linking vs Subscriptions
+
+| Use Case                   | Approach                                                    |
+|----------------------------|-------------------------------------------------------------|
+| **Process queue messages** | Use `queue.subscribe()` - creates Lambda triggered by queue |
+| **Send messages to queue** | Use `links=[queue]` - grants permissions to send messages   |
+| **Both read and write**    | Use both subscription AND link to another queue             |
```
michal-stlv: What's case for this? Sounds confusign, one lambda would send message to the queue and also process it?  how'd that work in stelvio code?

stelvio/aws/queue.py
```diff
@@ -0,0 +1,421 @@
...
+    "QueueConfigDict",
+]
+
+
+@dataclass(frozen=True, kw_only=True)
+class DlqConfig:
+    """Dead-letter queue configuration.
+
+    Args:
+        queue: Dead-letter queue - either a Queue component or queue name string.
```
michal-stlv: queue name works? isn't arn needed?

stelvio/aws/queue.py
```diff
@@ -0,0 +1,421 @@
...
+
+@dataclass(frozen=True, kw_only=True)
+class QueueConfig:
+    """Queue configuration."""
+
+    fifo: bool = False
+    delay: int = 0
+    visibility_timeout: int = 30
+    retention: int = 345600  # 4 days in seconds (default)
+    dlq: str | DlqConfig | DlqConfigDict | None = None
```
michal-stlv: if it's string you can't set retry?

stelvio/aws/queue.py
```diff
@@ -0,0 +1,421 @@
...
+                f"Dead-letter queue '{dlq_queue}' not found. "
+                "Make sure to create the DLQ before the main queue."
+            )
+        if not isinstance(queue_component, Queue):
+            raise TypeError(
+                f"Component '{dlq_queue}' is not a Queue, got {type(queue_component).__name__}"
+            )
+        return queue_component.arn
+
+    def _create_resources(self) -> QueueResources:
```
michal-stlv: I'd prefer to have this method before subscribe method.

stelvio/aws/queue.py
```diff
@@ -0,0 +1,421 @@
...
+
+        dlq_config = self.config.dlq
+        dlq_queue = dlq_config.queue
+
+        # If it's a Queue component, get its ARN directly
+        if isinstance(dlq_queue, Queue):
+            return dlq_queue.arn
+
+        # If it's a string, look up the queue by name
+        queue_component = ComponentRegistry.get_component_by_name(dlq_queue)
```
michal-stlv: Uf, this is hard to reason about for users I think. If we take string as queue name from it's stelvio component user will have hard time to know if that queue will be created before the main queue. I'd say let's just not support it by name? we could use arn potentially as that means user will probably use arn property of some queue(rather than building arn string). but then  they could pass arn of queue not creted by stelvio which should work and expands possibilities.

stelvio/aws/queue.py
michal-stlv: I'm wondering if we even want to have DlqConfig? 

It has just two properties and creates hassle for user (create dict or instance) and for us (all that handling code to parse it etc.). 

Maybe simple and better UX would be just to have `dlq: Queue | str | None` and then additional param `dlq_retry: int` param? If documented it's all good and simple?

stelvio/aws/queue.py
```diff
@@ -0,0 +1,421 @@
...
+            raise TypeError(
+                f"Component '{dlq_queue}' is not a Queue, got {type(queue_component).__name__}"
+            )
+        return queue_component.arn
+
+    def _create_resources(self) -> QueueResources:
+        # Build queue name (add .fifo suffix for FIFO queues)
+        queue_name = context().prefix(self.name)
+        if self.config.fifo:
+            queue_name = f"{queue_name}.fifo"
```
michal-stlv: What if it already ends with .fifo?

tests/aws/queue/test_queue.py
```diff
@@ -0,0 +1,854 @@
...
+        queues = [r for r in pulumi_mocks.created_resources if r.typ == "aws:sqs/queue:Queue"]
+        assert len(queues) == 2
+
+        # Find the main queue (not the DLQ)
+        main_queue_resource = next((q for q in queues if q.name == TP + "orders"), None)
+        assert main_queue_resource is not None
+
+        # Verify redrive policy is set
+        redrive_policy = main_queue_resource.inputs.get("redrivePolicy")
+        assert redrive_policy is not None
```
michal-stlv, line 526: same here, any specific value we can check?

tests/aws/queue/test_queue.py
```diff
@@ -0,0 +1,854 @@
+import json
```
michal-stlv, line 1: What if user creates multiple subscriptions with same handler but different batch sizes - what will or should happen? would be good to have test for that and document behavior. 

Also there's no test that creates a FIFO queue with a subscription AND verifies the EventSourceMapping is actually created correctly with the FIFO queue's ARN. but i guess it's ok as we have it for normal queue?

stelvio/aws/queue.py
sebst: If str, default value for retry should be set

stelvio/aws/queue.py
sebst: done

tests/aws/queue/test_queue.py
sebst: done

tests/aws/queue/test_queue.py
sebst: test added

docs/guides/queues.md
sebst: one lambda would receive a message and send a message to another queue

michal-stlv: hmm ok, wasn't clear to me from docs. maybe we can expand it there or change wording.

sebst: Clearer with this?
```
| **Pipeline (read → write)** | Subscribe to one queue, link to another for forwarding      |
```

sebst:
> Thoughts? Fine to merge as-is and add later, or do you want to include it now? (I'd still like to have it in 0.7)

follow up to

#125

docs/guides/queues.md
sebst: should be clearer now

michal-stlv (approved):

PR #115, opened 2025-12-31
feat: cron

michal-stlv:


docs/guides/cron.md
```diff
@@ -0,0 +1,225 @@
...
+| Minutes | 0-59 | , - * / |
+| Hours | 0-23 | , - * / |
+| Day-of-month | 1-31 | , - * ? / L W |
+| Month | 1-12 or JAN-DEC | , - * / |
+| Day-of-week | 1-7 or SUN-SAT | , - * ? L # |
+| Year | 1970-2199 | , - * / |
+
+!!! warning
+    All cron schedules run in UTC. Plan accordingly for your timezone.
+
```
sebst, line 70: maybe add a link to https://crontab.guru/

mkdocs.yml
```diff
@@ -70,6 +70,7 @@ nav:
       - Lambda Functions: guides/lambda.md
       - Dynamo DB: guides/dynamo-db.md
       - S3 Buckets: guides/s3.md
+      - Cron: guides/cron.md
```
sebst, line 73: also add to index.md

docs/guides/cron.md
```diff
@@ -0,0 +1,225 @@
...
+The simplest way to create a scheduled task is to provide a name, schedule expression, and handler:
+
+```python
+from stelvio.aws.cron import Cron
+
+# Run every hour
+Cron("hourly-cleanup", "rate(1 hour)", "tasks/cleanup.handler")
+
+# Run at 2 AM UTC daily
+Cron("nightly-report", "cron(0 2 * * ? *)", "tasks/report.handler")
```
sebst, line 16: Could we have something like 

```python
Cron("hourly-cleanup", rate("1 hour"), "tasks/cleanup.handler")
Cron("nightly-report", cron("0 2 * * ? *"), "tasks/report.handler")
```

maybe just expose functions `rate(str)` and `cron(str)` which for now could just be implemented like

```python
def rate(expr: str) -> str:
    return f"rate({expr})
```

We could use libs for units (like pint for rate or croniter for cron) later

sebst (changes requested):
Looks good. Guide needs to be added to `index.md`, too.
Otherwise some nitpicks

stelvio/aws/cron.py
```diff
@@ -0,0 +1,223 @@
...
+        _validate_cron_expression(schedule)
+    else:
+        raise ValueError(
+            f"Invalid schedule expression: must start with 'rate(' or 'cron(', got: {schedule}"
+        )
+
+
+def _parse_handler(
+    handler: str | FunctionConfig | FunctionConfigDict | Function | None,
+    opts: FunctionConfigDict,
```
sebst, line 75: should be function_opts

docs/guides/cron.md
michal-stlv: will that work? AWS doesn't use *nix cron syntax

docs/guides/cron.md
michal-stlv: yes and no. what will we gain? we'll need to import rate or cron functions but is there any benefit?

stelvio/aws/cron.py
michal-stlv: hmmm , kind of makes sense but not done in other places e.g. subscribe in dynamotable or route in api...

docs/guides/cron.md
sebst: > but is there any benefit?

Code looks less dependent on strings, could add libs for parsing units (e.g. pint)

docs/guides/cron.md
sebst: Yap, you're right, my mistake

docs/guides/cron.md
michal-stlv: what you mean less dependend on strings? you'd seel need to do string just without rate or cron prefix. and yuou need to import it. my vision is to have something really nice like "every 1 hour" , "each tuesday at 2pm" etc. etc. but for this first pass i'd say rate(something) and cron(somthing) as strings is ok and we will not gain anything if we make them functions -just users will have to import them.

sebst: Get your point. It's just a matter of preference. I'm fine with both

PR #119, opened 2026-01-09
feat: `Router` accepts routes only before accessing `resources`

sebst:


tests/aws/cloudfront/test_cloudfront_router.py
```diff
@@ -219,6 +219,20 @@ def test_add_route_rejects_function_with_url_config():
...
+    router = Router(name="test-router")
+    mock_bucket = Mock(spec=Bucket)
+
+    # Simulate that resources have been created by setting _resources
+    router._resources = Mock()
+
+    with pytest.raises(
+        RuntimeError, match="Cannot add routes after Router resources have been created."
+    ):
+        router.route("/static", mock_bucket)
```
michal-stlv: The current approach works, but it would be slightly more robust to actually trigger resource creation rather than directly setting the private _resources attribute. This would test the behavior through the public API and be resilient to any future changes in how resource creation state is tracked internally. (Mentioning this b/c i was thinking about other ways of triggering resource creation due to the problems with route here and api )

stelvio/aws/cloudfront/router.py
```diff
@@ -229,6 +229,8 @@ def route(
             component_or_url: The component (Bucket, Api, Function) or URL string to route to.
             function_url: Function URL config (only used if component_or_url is a Function).
         """
+        if self._resources is not None:
+            raise RuntimeError("Cannot add routes after Router resources have been created.")
```
michal-stlv, line 233: I'm wondering if we could use better exception type. I guess none is better from stdlib but maybe some custom exception type to better capture what's going on. Sadly I dont' have anything better on mind just have a feeling there's maybe better way.

sebst: Was thinking the same... Maybe for 1.0 we should have only `stelvio.*Exception`s. We could also have error codes and list them in the docs.

tests/aws/cloudfront/test_cloudfront_router.py
sebst: improved

stelvio/aws/cloudfront/router.py
michal-stlv: We do have some stelvio exceptions already. let's think in the background about a good name for this. It's easy change once we have a good name.

PR #128, opened 2026-01-15
feat: Filter configurations for `Queue`s

### LINKED ISSUE #125: Add filter options for Queue
The `Queue` component should support filters for the subscribe method.

The new signature should look like this:

```python
    def subscribe(
        self,
        name: str,
        handler: str | FunctionConfig | FunctionConfigDict | None = None,
        /,
        *,
        batch_size: int | None = None,
        filters: dict[str, Any]
        **opts: Unpack[FunctionConfigDict],
    ) -> QueueSubscription:
```

`QueueSubscription` should take the `filters` argument in its constructor.

Example for a filter:

```python
{
  "filters": [
    {
      "body": {
        "RequestType": ["Order"]
      }
    }
  ]
}
```

sebst:
closes #125

### DIFF
--- a/stelvio/aws/queue.py
+++ b/stelvio/aws/queue.py
@@ -8,7 +8,7 @@ from pulumi_aws.lambda_ import EventSourceMapping
 from pulumi_aws.sqs import Queue as SqsQueue
 
 from stelvio import context
-from stelvio.aws.function import Function, FunctionConfig, FunctionConfigDict
+from stelvio.aws.function import Function, FunctionConfig, FunctionConfigDict, parse_handler_config
 from stelvio.aws.permission import AwsPermission
 from stelvio.component import Component, link_config_creator, safe_name
 from stelvio.link import Link, LinkableMixin, LinkConfig
@@ -124,7 +124,7 @@ class QueueSubscription(Component[QueueSubscriptionResources]):
         self._check_filter_rules(filters)
         self.filters = filters
 
-        self.handler = self._create_handler_config(handler, opts)
+        self.handler = parse_handler_config(handler, opts)
 
     @staticmethod
     def _check_filter_rules(filters: list[SqsFilterDict] | None) -> None:
@@ -159,41 +159,6 @@ class QueueSubscription(Component[QueueSubscriptionResources]):
                     f"Each SQS message filter must be a dict, "
                     f"but filter at index {i} is {type(filter_rule).__name__}"
                 )
-
-    @staticmethod
-    def _create_handler_config(
-        handler: str | FunctionConfig | FunctionConfigDict | None,
-        opts: FunctionConfigDict,
-    ) -> FunctionConfig:
-        if isinstance(handler, dict | FunctionConfig) and opts:
-            raise ValueError(
-                "Invalid configuration: cannot combine complete handler "
-                "configuration with additional options"
-            )
-
-        if isinstance(handler, FunctionConfig):
-            return handler
-
-        if isinstance(handler, dict):
-            return FunctionConfig(**handler)
-
-        if isinstance(handler, str):
-            if "handler" in opts:
-                raise ValueError(
-                    "Ambiguous handler configuration: handler is specified both as positional "
-                    "argument and in options"
-                )
-            return FunctionConfig(handler=handler, **opts)
-
-        if handler is None:
-            if "handler" not in opts:
-                raise ValueError(
-                    "Missing handler configuration: when handler argument is None, "
-                    "'handler' option must be provided"
-                )
-            return FunctionConfig(**opts)
-
-        raise TypeError(f"Invalid handler type: {type(handler).__name__}")
 
     def _create_resources(self) -> QueueSubscriptionResources:
         # Create SQS link (mandatory for Lambda to poll SQS)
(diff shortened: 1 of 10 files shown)


stelvio/aws/queue.py
```diff
@@ -104,8 +119,47 @@ def __init__(
...
         self.batch_size = batch_size
+
+        # Validate and store filters
+        self._check_filter_rules(filters)
+        self.filters = filters
+
         self.handler = self._create_handler_config(handler, opts)
 
+    @staticmethod
+    def _check_filter_rules(filters: list[SqsFilterDict] | None) -> None:
```
michal-stlv, line 130: this could be just function

stelvio/aws/queue.py
```diff
@@ -14,9 +15,10 @@
 
 DEFAULT_QUEUE_BATCH_SIZE = 10
 DEFAULT_QUEUE_DELAY = 0
-DEFAULT_QUEUE_VISIBILITY_TIMEOUT = 30
+DEFAULT_QUEUE_VISIBILITY_TIMEOUT = 60
```
michal-stlv, line 18: just wondering why this change?

stelvio/aws/function/resources_codegen.py
```diff
@@ -46,7 +46,9 @@ def create_stlv_resource_file_content(
     if include_cors:
         lines.append("    cors: Final[CorsResource] = CorsResource()")
 
-    for link_name in link_properties_map:
+    for link_name, properties in link_properties_map.items():
+        if not properties:
+            continue
```
michal-stlv, line 51: this seems unrelated? is it fixing some bug?

tests/aws/queue/test_queue.py
```diff
@@ -899,6 +899,190 @@ def test_subscription_batch_size_validation_fifo_queue_invalid():
...
+def test_subscription_with_single_filter(pulumi_mocks, basic_queue):
+    """Test subscription with a single message filter."""
+    subscription = basic_queue.subscribe(
+        "filtered",
+        SIMPLE_HANDLER,
+        filters=[{"body": {"orderType": ["refund"]}}],
+    )
+
+    def check_filter(_):
+        mappings = [r for r in pulumi_mocks.created_resources if "EventSourceMapping" in r.typ]
```
michal-stlv: we should use `==` (with `"aws:lambda/eventSourceMapping:EventSourceMapping"`) instead of `in` and if used more then once create helper method in pulumi mocks.

tests/aws/queue/test_queue.py
```diff
@@ -899,6 +899,190 @@ def test_subscription_batch_size_validation_fifo_queue_invalid():
...
+        {"body": {"orderType": ["refund"]}},
+    ]
+    subscription = basic_queue.subscribe(
+        "multi-filtered",
+        SIMPLE_HANDLER,
+        filters=filters,
+    )
+
+    def check_filters(_):
+        import json
```
michal-stlv: Why inline import? And also it's already imported at the top of the file.

tests/aws/queue/test_queue.py
```diff
@@ -899,6 +899,190 @@ def test_subscription_batch_size_validation_fifo_queue_invalid():
...
+                },
+                "messageAttributes": {
+                    "region": ["us-west-2"],
+                },
+            }
+        ],
+    )
+
+    def check_complex_filter(_):
+        import json
```
michal-stlv: inline import of already imported json

tests/aws/queue/test_queue.py
```diff
@@ -899,6 +899,190 @@ def test_subscription_batch_size_validation_fifo_queue_invalid():
...
+        "filtered-batch",
+        SIMPLE_HANDLER,
+        filters=[{"body": {"priority": ["high"]}}],
+        batch_size=5,
+    )
+
+    def check_config(_):
+        mappings = [r for r in pulumi_mocks.created_resources if "EventSourceMapping" in r.typ]
+        assert len(mappings) == 1
+        mapping = mappings[0]
```
michal-stlv: these same 3 lines are repeated 5 times.

stelvio/aws/function/resources_codegen.py
sebst: This fixes a bug in the creation of resources with permissions but without properties

stelvio/aws/queue.py
sebst: This is to better align with defaults. Otherwise timeout of function is larger than visibility timeout by default

tests/aws/queue/test_queue.py
sebst: fixed

tests/aws/queue/test_queue.py
sebst: fixed

tests/aws/queue/test_queue.py
sebst: moved to pulumi mocks

tests/aws/queue/test_queue.py
sebst: moved to helper fn

michal-stlv (dismissed):

michal-stlv (approved):

PR #131, opened 2026-01-19
chore: pulumi and other updates

michal-stlv:


stelvio/_suppress_grpc.py
```diff
@@ -0,0 +1,21 @@
...
+
+os.environ["GRPC_VERBOSITY"] = "NONE"
+os.environ["GRPC_TRACE"] = ""
+os.environ["GRPC_ENABLE_FORK_SUPPORT"] = "0"
+
+_real_stderr_fd = os.dup(2)
+_devnull_fd = os.open(os.devnull, os.O_WRONLY)
+os.dup2(_devnull_fd, 2)
+os.close(_devnull_fd)
+sys.stderr = os.fdopen(_real_stderr_fd, "w", buffering=1)
```
sebst, line 21: Should that run at import time? any way to encapsulate that into a function?

michal-stlv: yeah needs to run before anything else. hopefully it's temp fix until grpc bug is fixed.

PR #147, opened 2026-02-09
fix: create acm certs in `us-east-1`

### LINKED ISSUE #146: ACM certificates are created in local region for cloudfront
ACM for cloudfront need to be created in us-east-1 regardless of the region used for other components

sebst:
closes #146

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.7.1b2 (2026-02-DD)
+## 0.7.1b2 (2026-02-20)
 
 This is a bug-fix release.
 


tests/aws/s3/test_cloudfront.py
```diff
@@ -60,6 +60,23 @@ def app_context_with_dns(mock_dns):
     _ContextStore.clear()
 
 
+@pytest.fixture
+def app_context_with_dns_eu_west(mock_dns):
```
michal-stlv: we have this same fixture 3 times?

stelvio/aws/acm.py
```diff
@@ -34,8 +34,10 @@ def __init__(
         name: str,
         domain_name: str,
         customize: AcmValidatedDomainCustomizationDict | None = None,
+        provider: pulumi_aws.Provider | None = None,
```
michal-stlv: maybe accept `region: str | None = None` instead of a raw `pulumi_aws.Provider`. That way AcmValidatedDomain can create the provider internally when a non-default region is needed, and callers just pass `region="us-east-1"` instead of each constructing their own Provider. This would also eliminate the duplicated provider creation code in cloudfront.py, router.py, and api.py.

tests/aws/cloudfront/test_cloudfront_router_pulumi.py
```diff
@@ -48,6 +48,32 @@ def app_context_with_dns():
...
+    yield mock_dns
+    _ContextStore.clear()
+    _ContextStore.set(
+        AppContext(
+            name="test",
+            env="test",
+            aws=AwsConfig(profile="default", region="us-east-1"),
+            home="aws",
+        )
+    )
```
michal-stlv: The other two `app_context_with_dns_eu_west` fixtures (in test_custom_domain.py and test_cloudfront.py) just call (_ContextStore.clear()` after yield. This one adds an extra `_ContextStore.set(...)` that leaves a stale context (us-east-1, no DNS) for subsequent tests. Should be removed to match the others. Or is there reason for this?

stelvio/aws/cloudfront/cloudfront.py
```diff
@@ -73,10 +73,16 @@ def _create_resources(self) -> CloudFrontDistributionResources:
...
+            us_east_1_provider = pulumi_aws.Provider(
+                context().prefix(f"{self.name}-us-east-1-provider"),
+                region="us-east-1",
+            )
             acm_validated_domain = AcmValidatedDomain(
                 f"{self.name}-acm-validated-domain",
                 domain_name=self.custom_domain,
                 customize=self._customize.get("acm_validated_domain"),
+                provider=us_east_1_provider,
             )
```
michal-stlv: maybe we could skip the us-east-1 provider creation when the user's configured region is already us-east-1 — no need to create a redundant  provider. Would be good to add a test for that case too.

tests/aws/s3/test_cloudfront.py
sebst: fixed

stelvio/aws/acm.py
sebst: changed

tests/aws/cloudfront/test_cloudfront_router_pulumi.py
sebst: done

stelvio/aws/cloudfront/cloudfront.py
sebst: done

PR #162, opened 2026-02-28
feat: appsync component

### LINKED ISSUE #150: AppSync component — API design
This issue defines the API design for the AppSync component. The goal is a Pythonic, Stelvio-consistent interface that makes the common case (Lambda + PowerTools) simple while supporting all AppSync features.

## Quick Example

The most common pattern — Lambda resolvers with PowerTools:

```python
api = AppSync("myapi", "schema.graphql",
    auth=CognitoAuth(user_pool_id="..."),
    additional_auth=["iam", ApiKeyAuth()],
)

posts = api.data_source_lambda("posts", handler="resolvers/posts.handler", links=[table])
api.query("getPost", posts)
api.query("listPosts", posts)
api.mutation("createPost", posts)
```

One Lambda handles all three resolvers (PowerTools router pattern). The data source is defined once, resolvers just reference it. AppSync uses Direct Lambda Resolver — no JS mapping code needed.

---

## Constructor

```python
AppSync(
    name: str,
    schema: str,                # file path if it exists, otherwise inline SDL string
    *,
    auth: AuthConfig,
    additional_auth: list[AuthConfig] | None = None,
    domain: str | None = None,  # custom domain (ACM + DNS auto-handled, same pattern as Api)
    customize: AppSyncCustomizationDict | None = None,  # keys: "api", "domain_name"
)
```

**Properties:**
- `url: Output[str]` — GraphQL endpoint URL
- `arn: Output[str]` — API ARN
- `api_id: Output[str]` — API ID
- `api_key: Output[str] | None` — API key value, or None if API_KEY auth isn't configured

---

## Auth

`AuthConfig` is a union type:

```python
AuthConfig = Literal["iam"] | ApiKeyAuth | CognitoAuth | OidcAuth | LambdaAuth
```

Five auth modes. `"iam"` is a string literal (no config needed). Everything else is a frozen dataclass:

```python
@dataclass(frozen=True)
class ApiKeyAuth:
    expires: int = 365  # days, AWS max 365

@dataclass(frozen=True)
class CognitoAuth:
    user_pool_id: str
    region: str | None = None
    app_id_client_regex: str | None = None

@dataclass(frozen=True)
class OidcAuth:
    issuer: str
    client_id: str | None = None
    auth_ttl: int | None = None
    iat_ttl: int | None = None

@dataclass(frozen=True, kw_only=True)
class LambdaAuth:
    handler: str | FunctionConfig | Function
    result_ttl: int | None = None
    identity_validation_expression: str | None = None
    # plus **fn_opts (Unpack[FunctionConfigDict]) for links, memory, timeout, etc.
```

`LambdaAuth` creates a Lambda authorizer. It accepts `handler` as `str | FunctionConfig | Function` plus `**fn_opts` for links, memory, timeout — same pattern as `data_source_lambda`. The authorizer Lambda might need database access for token validation:

```python
auth=LambdaAuth(handler="resolvers/auth.handler", links=[users_table], memory=256)
```

### Multi-auth

Set one default + optional additional modes. Per-field control uses schema directives.

```python
api = AppSync("myapi", "schema.graphql",
    auth=CognitoAuth(user_pool_id="..."),       # default for all fields
    additional_auth=["iam", ApiKeyAuth()],       # enables @aws_iam and @aws_api_key in schema
)
```

A directive (`@aws_iam`, `@aws_api_key`, `@aws_cognito_user_pools`, `@aws_oidc`) can only be used if that auth mode is configured (default or additional). Fields without directives use the default.

**Gotcha worth documenting:** if a return type is used by both Cognito and IAM callers, it needs both directives on the type itself (`@aws_iam @aws_cognito_user_pools`).

When API_KEY auth is configured, Stelvio auto-creates the ApiKey resource and exposes the value via `api.api_key`.

---

## Data Sources

Each data source type has its own method with type-specific params. Each method returns an `AppSyncDataSource` object that you pass to resolver methods. It has a `.resources` property:

```python
@dataclass(frozen=True)
class AppSyncDataSourceResources:
    data_source: appsync.DataSource
    service_role: iam.Role
    function: Function | None = None  # only set for Lambda data sources
```

```python
posts  = api.data_source_lambda("posts", handler="resolvers/posts.handler", links=[table])
items  = api.data_source_dynamo("items", table=items_table)
ext    = api.data_source_http("ext", url="https://api.example.com")
db     = api.data_source_rds("db", cluster_arn="...", secret_arn="...", database="mydb")
search = api.data_source_opensearch("search", endpoint="https://...")
```

| Method | Key params |
|--------|-----------|
| `data_source_lambda(name, handler, *, links, customize, **fn_opts)` | handler: str / FunctionConfig / Function |
| `data_source_dynamo(name, *, table, customize)` | table: Stelvio DynamoDB component |
| `data_source_http(name, *, url, customize)` | url: base URL string |
| `data_source_rds(name, *, cluster_arn, secret_arn, database, customize)` | Aurora Data API connection info |
| `data_source_opensearch(name, *, endpoint, customize)` | OpenSearch domain endpoint |

NONE has no method — pass `None` directly as the data source in resolver methods.

For `data_source_dynamo`, v1 requires a Stelvio DynamoDB component — no raw ARN strings.

Stelvio auto-creates an IAM service role (trusted by `appsync.amazonaws.com`) per data source with the permissions that type needs. All data source methods accept `customize={"data_source": {...}, "service_role": {...}}`.

---

## Resolvers

Four methods — `query()` and `mutation()` are shortcuts for the most common types. `resolver()` handles nested types. `subscription()` is for enhanced server-side filtering (basic subscriptions just use `@aws_subscribe` in the schema and need no infra code).

```python
api.query(field, data_source, *, code=None, customize=None)
api.mutation(field, data_source, *, code=None, customize=None)
api.subscription(field, data_source, *, code=None, customize=None)
api.resolver(type, field, data_source, *, code=None, customize=None)
```

- **data_source** — required positional arg. Pass a data source object, `None` for NONE, or a `list[PipeFunction]` for pipeline resolvers.
- **code** — APPSYNC_JS code as inline string or `.js` file path.
- **customize** — `{"resolver": {...}}` for things like caching_config, max_batch_size, sync_config.

Each method returns an `AppSyncResolver` object with `.resources` exposing the underlying `appsync.Resolver`.

### When is code needed?

Depends on the data source type:

- **Lambda** — no code needed. Direct Lambda Resolver auto-forwards the full GraphQL context to the Lambda. The Lambda handles everything.
- **DynamoDB** — code required. The JS tells AppSync which operation to perform (GetItem, Query, etc.) and how to map args to keys.
- **HTTP, RDS, OpenSearch** — code required. The JS specifies the request (HTTP method/path, SQL query, search query).
- **NONE** — code optional. Without it, Stelvio auto-generates a passthrough that forwards `ctx.args` as the result. With it, you provide custom JS (e.g., timestamp injection, data enrichment).

```python
# Lambda — no code
api.query("getPost", posts)

# DynamoDB — JS required
api.query("getItem", items, code="resolvers/getItem.js")

# HTTP — JS required
api.query("fetchExt", ext, code="resolvers/fetchExt.js")

# NONE — auto passthrough
api.mutation("sendMessage", None)

# NONE — custom code
api.mutation("enriched", None, code="resolvers/enrich.js")
```

### Validation

- Missing `code=` on DynamoDB/HTTP/RDS/OpenSearch data source → clear error
- Duplicate type+field → error

---

## Pipeline Resolvers

Pipelines chain multiple steps in sequence. Each step is an AppSync Function (confusing AWS name — it's a pipeline step, not a Lambda). Each step has its own data source and JS code. Steps read the previous result via `ctx.prev.result` and share data via `ctx.stash`.

```python
auth_step   = api.pipe_function("checkAuth", None, code="resolvers/auth.js")
delete_step = api.pipe_function("doDelete", items, code="resolvers/delete.js")
api.mutation("deletePost", [auth_step, delete_step])
```

Common pattern: auth check (NONE data source, just inspects `ctx.identity`) → business logic (real data source).

### pipe_function

```python
api.pipe_function(name, data_source, *, code, customize=None)
```

- **code** — required, always JS (inline or file path)
- **data_source** — which backend this step talks to, `None` for NONE
- **customize** — `{"function": {...}}`

Returns a `PipeFunction` object with `.resources` exposing the underlying `appsync.Function`. Works with any resolver type.

---

## NONE Data Source

NONE means "no backend call." Pass `None` as the data source.

When no `code=` is provided, Stelvio generates a passthrough:
```javascript
export function request(ctx) { return { payload: ctx.args }; }
export function response(ctx) { return ctx.result; }
```

The main use case is **real-time pub/sub without persistence**. A mutation with NONE just passes args through as the "result." Any `@aws_subscribe` subscription watching that mutation receives it automatically. No database, no Lambda — pure broadcast.

For custom behavior, provide `code=` (timestamp injection, identity enrichment, subscription filtering, etc.).

This needs clear documentation — explain what passthrough does, the pub/sub pattern, and when to use NONE vs Lambda.

---

## Runtime

APPSYNC_JS only (JavaScript). VTL is legacy and we don't actively support it. If someone really needs VTL, they can use `customize={"resolver": {"request_template": "..."}}`.

---

## Link System

**AppSync as Linkable** — other Lambdas can link to the API (e.g., backend services calling mutations via IAM-signed HTTP):

```python
notifier = Function("notifier", handler="jobs/notify.handler", links=[api])
# → STLV_MYAPI_URL env var
# → STLV_MYAPI_API_KEY env var (if API_KEY auth configured)
# → appsync:GraphQL permission on {arn}/*
```

**Lambda data source links** — standard function config, nothing AppSync-specific:

```python
posts = api.data_source_lambda("posts", handler="resolvers/posts.handler", links=[table, queue])
```

---

## Customization

Standard `customize=` escape hatch on constructor, data source methods, resolver methods, and `pipe_function()`. Same pattern as Topic.subscribe().

---

## Code Generation Helpers (nice-to-have)

Optional utility functions that generate APPSYNC_JS strings for common operations, so Python devs don't have to write JS for simple cases:

```python
from stelvio.aws.appsync import dynamo_get, dynamo_scan, dynamo_remove

api.query("getItem", items, code=dynamo_get("id"))
api.query("listItems", items, code=dynamo_scan())
api.mutation("deleteItem", items, code=dynamo_remove("id"))
api.query("getByKeys", items, code=dynamo_get(pk="userId", sk="postId"))
```

Pure functions returning JS strings. `code=` stays the universal param. Could extend to other data source types (RDS basic queries, HTTP GET/POST patterns) over time.

---

## Docs Should Cover

- PowerTools as the recommended Lambda pattern (show Router for splitting across files)
- Data source concept — what's infra config vs resolver code, why data sources are explicit
- DynamoDB direct resolvers — when to use vs Lambda
- Pipeline resolvers — auth-before-action pattern
- Subscriptions — basic (`@aws_subscribe`, no infra) vs enhanced filtering
- NONE data source — passthrough, pub/sub pattern, when to use
- All data source types with JS code examples
- Auth — modes, per-field directives, multi-auth, shared return type gotcha
- Linking — calling AppSync from backend Lambdas via IAM

sebst:
closes #150 

https://x.com/twitter/status/2028779923712758153

### DIFF
--- a/docs/guides/appsync.md
+++ b/docs/guides/appsync.md
@@ -828,6 +828,7 @@ api.query("getPost", posts, customize={
     To apply customizations to all instances of a component type, use the `customize` option in `StelvioAppConfig` with **component classes** as keys:
 
     ```python
+    from stelvio.config import StelvioAppConfig
     from stelvio.aws.appsync import AppSync, AppSyncDataSource
 
     @app.config


docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+## Schema
+
+The `schema` parameter accepts either a file path (relative to project root) or an inline SDL string:
+
+```python
+# File path — reads from project root
+api = AppSync("myapi", "schema.graphql", auth=...)
+
+# Inline SDL string
+api = AppSync("myapi", """
```
michal-stlv: This sample schema is missing type Post so it's invalid.

docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+```
+
+Stelvio treats values ending in `.graphql` or `.gql` as file paths. If that file is missing, it raises `FileNotFoundError`. Other values are treated as inline SDL strings.
+
+## Component Lifecycle
+
+`AppSync`, `AppSyncDataSource`, `AppSyncResolver`, and `PipeFunction` are all Stelvio components with lazy resource creation.
+
+- Builder methods (`data_source_*`, `query`/`mutation`/`subscription`/`resolver`, `pipe_function`) register child components.
+- Actual Pulumi resources are created when component resources are resolved during `drive()`.
```
michal-stlv: user's don't know what is drive(), that's internal detail. just say during deployment.../deploy

docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+- Child components always resolve parent API details from `api.resources` rather than relying on parent-side mutation hooks.
```
michal-stlv: how's this useful to the users?

docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+| `handler`                         | —       | Handler as string, `FunctionConfig`, or `Function` (required) |
+| `result_ttl`                      | `None`  | Authorization result cache TTL in seconds        |
+| `identity_validation_expression`  | `None`  | Regex to validate the authorization token        |
+| `links`                           | `[]`    | Resources to link to the authorizer function     |
+| `memory`                          | `None`  | Memory size in MB                                |
+| `timeout`                         | `None`  | Timeout in seconds                               |
+| `environment`                     | `{}`    | Environment variables                            |
+
+!!! info "Function options"
+    The `links`, `memory`, `timeout`, and `environment` fields are convenience options for when `handler` is a string. If you pass a `FunctionConfig` or `Function` instance, configure those on the handler directly.
```
michal-stlv: It shouldn't be necessary to mention all of this here, instead just refer to Function docs, this is common pattern everywhere, we don't need to re-document it.

docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+from stelvio.aws.dynamo_db import DynamoTable, AttributeType
+
+items_table = DynamoTable("items",
+    fields={"id": AttributeType.STRING},
+    partition_key="id",
+)
+
+items = api.data_source_dynamo("items", table=items_table)
+
+# DynamoDB resolvers require JS code
```
michal-stlv, line 293: Don't we have these helper functions? (they were in previous PR), if so we should mention them here.

docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+| Data Source   | Code Required? | Reason |
+|---------------|---------------|--------|
+| **Lambda**    | No            | Direct Lambda Resolver forwards the full GraphQL context to your Lambda |
+| **DynamoDB**  | Yes           | JS tells AppSync which operation to perform (GetItem, Query, etc.) |
+| **HTTP**      | Yes           | JS specifies the HTTP method, path, and response mapping |
+| **RDS**       | Yes           | JS provides the SQL query |
+| **OpenSearch** | Yes          | JS specifies the search query |
+| **NONE**      | No (optional) | Without code, Stelvio generates a passthrough; with code, you provide custom JS |
+
+The `code` parameter accepts either an inline JavaScript string or a `.js` file path (relative to project root):
```
michal-stlv, line 420: What is this referring to? you don't mention any code param.

docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+Whether you need to provide `code=` depends on the data source type:
+
+| Data Source   | Code Required? | Reason |
+|---------------|---------------|--------|
+| **Lambda**    | No            | Direct Lambda Resolver forwards the full GraphQL context to your Lambda |
+| **DynamoDB**  | Yes           | JS tells AppSync which operation to perform (GetItem, Query, etc.) |
+| **HTTP**      | Yes           | JS specifies the HTTP method, path, and response mapping |
+| **RDS**       | Yes           | JS provides the SQL query |
+| **OpenSearch** | Yes          | JS specifies the search query |
+| **NONE**      | No (optional) | Without code, Stelvio generates a passthrough; with code, you provide custom JS |
```
michal-stlv: > Without code, Stelvio generates a passthrough; with code, you provide custom JS

This is confusing. If it's None then is custom js valid actually?

docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+    handler="resolvers/posts.handler",
+    links=[table, queue],
+)
+```
+
+| Param        | Description                                             |
+|--------------|---------------------------------------------------------|
+| `name`       | Data source name (unique within this API)               |
+| `handler`    | Handler as string, `FunctionConfig`, or `Function`      |
+| `links`      | Resources to link to the Lambda function                |
```
michal-stlv: hmmm how's this working? links is part of the functionconfig no? so no need for it to be separate arg here?

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+from stelvio.aws.permission import AwsPermission
+from stelvio.component import Component, link_config_creator, safe_name
+from stelvio.dns import DnsProviderNotConfiguredError, Record
+from stelvio.link import Link, Linkable, LinkableMixin, LinkConfig
+
+_DS_TYPES_REQUIRING_CODE = (DS_TYPE_DYNAMO, DS_TYPE_HTTP, DS_TYPE_RDS, DS_TYPE_OPENSEARCH)
+
+
+def _validate_no_duplicate_auth(auth: AuthConfig, additional_auth: list[AuthConfig]) -> None:
+    all_types = [_auth_type_string(auth)]
```
michal-stlv: nitpick but could be set?

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+    raise TypeError(f"Unexpected auth config type: {type(auth).__name__}")
+
+
+def _build_cognito_config(auth: CognitoAuth) -> dict[str, Any]:
+    config: dict[str, Any] = {"user_pool_id": auth.user_pool_id}
+    if auth.region:
+        config["aws_region"] = auth.region
+    if auth.app_id_client_regex:
+        config["app_id_client_regex"] = auth.app_id_client_regex
+    return config
```
michal-stlv: could be method on CognitoAuth dataclass?

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+
+def _build_oidc_config(auth: OidcAuth) -> dict[str, Any]:
+    config: dict[str, Any] = {"issuer": auth.issuer}
+    if auth.client_id:
+        config["client_id"] = auth.client_id
+    if auth.auth_ttl is not None:
+        config["auth_ttl"] = auth.auth_ttl
+    if auth.iat_ttl is not None:
+        config["iat_ttl"] = auth.iat_ttl
+    return config
```
michal-stlv: could be method on OidcAuth dataclass?

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+    return provider
+
+
+def _build_lambda_authorizer_config(auth: LambdaAuth, invoke_arn: Output[str]) -> dict[str, Any]:
+    config: dict[str, Any] = {"authorizer_uri": invoke_arn}
+    if auth.result_ttl is not None:
+        config["authorizer_result_ttl_in_seconds"] = auth.result_ttl
+    if auth.identity_validation_expression:
+        config["identity_validation_expression"] = auth.identity_validation_expression
+    return config
```
michal-stlv: could be mehtod on LambdaAuth dataclass?

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+class AppSync(Component[AppSyncResources, AppSyncCustomizationDict], LinkableMixin):
+    def __init__(  # noqa: PLR0913
+        self,
+        name: str,
+        schema: str,
+        *,
+        auth: AuthConfig,
+        additional_auth: list[AuthConfig] | None = None,
+        domain: str | None = None,
+        customize: AppSyncCustomizationDict | None = None,
```
michal-stlv: since we hit PLR0913, worth using AppSyncConfig? (with dict version too)

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+        if self.resources.api_key is None:
+            return None
+        return self.resources.api_key.key
+
+    def data_source_lambda(
+        self,
+        name: str,
+        handler: str | FunctionConfig | Function,
+        *,
+        links: list[Link | Linkable] | None = None,
```
michal-stlv: why extra link? part of FunctionConfig no?

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+        return None
+
+    def _create_resources(self) -> AppSyncResources:
+        prefix = context().prefix
+
+        auth_function, additional_auth_functions = self._create_auth_lambdas()
+        api_args = self._build_api_args(auth_function, additional_auth_functions)
+
+        graphql_api = appsync.GraphQLApi(
+            prefix(self.name),
```
michal-stlv: safe_name?

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+            type=DS_TYPE_NONE,
+            opts=self._resource_opts(),
+        )
+
+        # Temporarily set _resources so children can access parent resources
+        self._resources = AppSyncResources(
+            api=graphql_api,
+            api_key=api_key_resource,
+            auth_permissions=auth_permissions,
+        )
```
michal-stlv: This breaks patter of all components. self._resources is not for temp stuff. it's final and immutable. we never write there directly it's handled by base component object that writes there return value of this method.

stelvio/aws/appsync/data_source.py
```diff
@@ -0,0 +1,367 @@
...
+    @property
+    def api(self) -> "AppSync":
+        return self._api
+
+    @property
+    def ds_type(self) -> str:
+        return self._ds_type
+
+    @property
+    def api_name(self) -> str | None:
```
michal-stlv: it's never None . Or is it?

stelvio/aws/appsync/data_source.py
```diff
@@ -0,0 +1,367 @@
...
+class AppSyncDataSource(
+    Component[AppSyncDataSourceResources, "AppSyncDataSourceCustomizationDict"]
+):
+    """A data source registered with an AppSync API.
+
+    Created by AppSync builder methods (data_source_lambda, data_source_dynamo, etc.).
+    Pass to resolver methods (query, mutation, etc.) to wire resolvers to data sources.
+    """
+
+    def __init__(  # noqa: PLR0913
```
michal-stlv: lot of params. AppSyncDataSourceConfig?

stelvio/aws/appsync/resolver.py
```diff
@@ -0,0 +1,226 @@
...
+        self,
+        api: "AppSync",
+        name: str,
+        data_source: "AppSyncDataSource | None",
+        *,
+        code: str,
+        customize: "AppSyncPipeFunctionCustomizationDict | None" = None,
+    ) -> None:
+        super().__init__("stelvio:aws:PipeFunction", f"{api.name}-fn-{name}", customize=customize)
+        self._name = name
```
michal-stlv: this mutate name after init? 

i think you need to create separate pipe_function_name field to store just name to use later. this _name is global pulumi identifier. (you did this in data source with _data_source_name.

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+        config["identity_validation_expression"] = auth.identity_validation_expression
+    return config
+
+
+_VALID_DS_CUSTOMIZE_KEYS = frozenset({"data_source", "service_role"})
+_VALID_RESOLVER_CUSTOMIZE_KEYS = frozenset({"resolver"})
+_VALID_PIPE_FN_CUSTOMIZE_KEYS = frozenset({"function"})
+
+
+def _validate_customize_keys(
```
michal-stlv: we have this in base class already?

tests/aws/appsync/test_appsync_codegen.py
```diff
@@ -0,0 +1,192 @@
...
+)
+
+
+def test_dynamo_get_single_key():
+    code = dynamo_get("id")
+    assert "operation: 'GetItem'" in code
+    assert "id: ctx.args.id" in code
+    assert "util.dynamodb.toMapValues" in code
+    assert "export function request" in code
+    assert "export function response" in code
```
michal-stlv: here and all other places, we want more precise assertion than just assert "key: ctx.args.id" in code. we are here for specific template so it needs to be precise.

tests/aws/appsync/test_appsync_resolvers.py
```diff
@@ -0,0 +1,476 @@
...
+  return { payload: { ...ctx.args, sentAt: util.time.nowISO8601() } };
+}
+export function response(ctx) { return ctx.result; }
+"""
+    api.mutation("sendMessage", None, code=custom_js)
+
+    def check_resources(_):
+        resolvers = pulumi_mocks.created_appsync_resolvers()
+        assert len(resolvers) == 1
+        assert "sentAt" in resolvers[0].inputs["code"]
```
michal-stlv: should be == custom_js ? to be precise?

tests/aws/appsync/test_appsync_data_sources.py
```diff
@@ -0,0 +1,591 @@
...
+def test_lambda_data_source_resources_accessible(pulumi_mocks, project_cwd):
+    api = make_api()
+    posts = api.data_source_lambda("posts", handler="functions/simple.handler")
+    api.query("getPost", posts)
+
+    def check_resources(_):
+        assert posts.resources is not None
+        assert posts.resources.data_source is not None
+        assert posts.resources.service_role is not None
+        assert posts.resources.function is not None
```
michal-stlv: better assertions needed here. not just not None. what it should be?

stelvio/aws/appsync/resolver.py
```diff
@@ -0,0 +1,226 @@
...
+    def data_source(self) -> "AppSyncDataSource | list[PipeFunction] | None":
+        return self._data_source
+
+    @property
+    def code(self) -> str | None:
+        return self._code
+
+    @property
+    def customize(self) -> "AppSyncResolverCustomizationDict":
+        return self._customize
```
michal-stlv: what's this?

stelvio/aws/appsync/resolver.py
```diff
@@ -0,0 +1,226 @@
...
+    def field_name(self) -> str:
+        return self._field_name
+
+    @property
+    def data_source(self) -> "AppSyncDataSource | list[PipeFunction] | None":
+        return self._data_source
+
+    @property
+    def code(self) -> str | None:
+        return self._code
```
michal-stlv: why's this neede? what's use case?

stelvio/aws/appsync/resolver.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+@final
+class AppSyncResolver(Component[AppSyncResolverResources, "AppSyncResolverCustomizationDict"]):
+    """A resolver registered with an AppSync API.
+
+    Created by AppSync resolver methods (query, mutation, subscription, resolver).
+    """
+
+    def __init__(  # noqa: PLR0913
```
michal-stlv: AppSyncResolverConfig since we hit PLR0913?

tests/aws/appsync/test_appsync_custom_domain.py
```diff
@@ -0,0 +1,125 @@
...
+        domain_cname_records = [
+            r
+            for r in dns_records
+            if r.inputs.get("type") == "CNAME" and r.inputs.get("name") == "api.example.com"
+        ]
+        assert len(domain_cname_records) == 1
+        assert (
+            domain_cname_records[0]
+            .inputs["content"]
+            .endswith(".appsync-api.us-east-1.amazonaws.com")
```
michal-stlv: should be exact value since mock returns deterministic f"{resource_id}.appsync-api.{region}.amazonaws.com"

tests/integration/test_appsync.py
```diff
@@ -0,0 +1,203 @@
...
+        api.query("echo", ds)
+
+    outputs = stelvio_env.deploy(infra)
+
+    assert_appsync_api(
+        outputs["appsync_key-api_id"],
+        authentication_type="API_KEY",
+    )
+    assert "appsync_key-api_api_key" in outputs
+    assert outputs["appsync_key-api_api_key"]
```
michal-stlv: should be assert outputs["appsync_key-api_api_key"].startswith("da2-") since real AWS API keys always start with da2-

tests/integration/test_appsync.py
```diff
@@ -0,0 +1,203 @@
...
+    outputs = stelvio_env.deploy(infra)
+
+    assert_appsync_api(
+        outputs["appsync_multi-auth-api_id"],
+        authentication_type="AWS_IAM",
+        additional_auth_count=1,
+        additional_auth_types=["API_KEY"],
+    )
+    assert "appsync_multi-auth-api_api_key" in outputs
+    assert outputs["appsync_multi-auth-api_api_key"]
```
michal-stlv: should be assert outputs["appsync_key-api_api_key"].startswith("da2-") since real AWS API keys always start with da2-

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+            if dns is None:
+                raise DnsProviderNotConfiguredError(
+                    "DNS provider is not configured. "
+                    "Please set up a DNS provider to use custom domains."
+                )
+
+            acm_validated_domain = acm.AcmValidatedDomain(
+                f"{self.name}-acm-domain",
+                **self._customizer("acm_domain", {"domain_name": self._domain}),
+            )
```
michal-stlv: _customizer is for merging args that go directly to a Pulumi resource constructor. When creating a child Stelvio component (like AcmValidatedDomain or Function), the established pattern is to pass the customize sub-dict directly as the customize= kwarg: customize=self._customize.get("acm_domain"). See QueueSubscription, TopicSubscription, Cron, BucketNotifySubscription — all pass self._customize.get("function") to Function(). Don't spread _customizer output into a Component constructor.

docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+```python
+from stelvio.aws.appsync import dynamo_get, dynamo_scan, dynamo_put, dynamo_remove, dynamo_query
+
+api.query("getItem", items, code=dynamo_get("id"))
+api.query("listItems", items, code=dynamo_scan())
+api.mutation("createItem", items, code=dynamo_put())
+api.mutation("deleteItem", items, code=dynamo_remove("id"))
+```
+
+### Available Helpers
```
michal-stlv, line 627: The docs for dynamo_put and dynamo_scan helpers need more explanation:

dynamo_put(): Explain that all mutation arguments become item attributes, and without key_fields an auto-generated id is used as the partition key. With key_fields, those args are extracted for the DynamoDB key and all args (including keys) become attributes.

dynamo_scan(): Explain that limit caps items per page, and next_token_arg is the GraphQL argument name clients pass to request the next page (AppSync returns a pagination token in the response, clients send it back in the next query). Default is "nextToken" — the GraphQL schema needs a matching argument.

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+@final
+@dataclass(frozen=True)
+class AppSyncResources:
+    api: appsync.GraphQLApi
+    api_key: appsync.ApiKey | None
+    auth_permissions: list[lambda_.Permission] | None = None
+    acm_domain: acm.AcmValidatedDomain | None = None
+    custom_domain: appsync.DomainName | None = None
+    domain_association: appsync.DomainNameApiAssociation | None = None
+    domain_dns_record: Record | None = None
```
michal-stlv, line 81: custom_domain key in AppSyncCustomizationDict (config.py:152) is dead — it's defined in the TypedDict but never passed to _customizer("custom_domain", ...) anywhere in the code. Users setting customize={"custom_domain": {...}} would get silently ignored. Either remove it, or if it was meant to customize the appsync.DomainName resource, that's already handled by the domain_name key (appsync.py:617).

Also, `acm_domain`, `domain_association`, and `domain_dns_record` are typed as `Any | dict[str, Any] | None` — the other keys all use proper Pulumi Args types (e.g., GraphQLApiArgs, DataSourceArgs, RoleArgs). `domain_association` should be `DomainNameApiAssociationArgs | dict[str, Any] | None`. `domain_dns_record` is cross-provider (Route53 or Cloudflare) so `dict[str, Any] | None` is reasonable there. `acm_domain` depends on resolving the _customizer vs pass-through question — if it stays as a pass-through to AcmValidatedDomain.customize, the type should be `AcmValidatedDomainCustomizationDict | None`.

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+
+    def _get_api_key_auth(self) -> ApiKeyAuth | None:
+        if isinstance(self._auth, ApiKeyAuth):
+            return self._auth
+        for auth in self._additional_auth:
+            if isinstance(auth, ApiKeyAuth):
+                return auth
+        return None
+
+    def _create_resources(self) -> AppSyncResources:
```
michal-stlv, line 450: `_create_resources()` should not set `self._resources` mid-execution (line 576-581) or trigger child resource creation via the for-loops (lines 583-599). This breaks the Component pattern contract — `_create_resources()` should only create its own resources and return them. The base class handles setting `_resources` from the return value.

Follow the DynamoSubscription pattern instead. Look at how `DynamoSubscription._create_resources()` accesses `self.table.stream_arn` — this triggers `DynamoTable._create_resources()` via lazy eval through the `.resources` property. No parent manually triggers children. Stelvio's `drive()` calls `.resources` on every registered component.

**What should happen:**

1. `AppSync._create_resources()` creates only its own resources (GraphQLApi, ApiKey, NONE DataSource, domain) and returns `AppSyncResources`. No for-loops. No `self._resources = ...` hack.

2. Each child accesses `self._api.resources.api.id` in its own `_create_resources()`. This triggers the parent's lazy creation if it hasn't happened yet — exactly like `self.table.stream_arn` triggers DynamoTable's creation.

3. `drive()` calls `.resources` on all registered components. Since children call `super().__init__()` they're in the ComponentRegistry, so `drive()` will trigger them automatically.

4. Each child does its own `pulumi.export()` in its own `_create_resources()`.

Ordering is safe regardless of which component `drive()` visits first:
- **Child first**: `self._api.resources.api.id` → triggers parent → parent creates API + NONE DS + domain → returns → child continues with its own resources
- **Parent first**: creates its own resources, returns. Children triggered later, `self._api.resources` returns cached result.

**What goes away:**
- Lines 576-581: the `self._resources = AppSyncResources(...)` temporary assignment
- Lines 583-599: the for-loops triggering children and doing their exports
- The `if self._api._resources is None` guard checks in data_source.py:164, resolver.py:87, resolver.py:189 — replaced by normal lazy access via `self._api.resources`
- The 3 `noqa: SLF001` comments on those guards

Also: extract domain creation into a `_create_domain_resources()` helper — the method is too long and domain logic is self-contained.

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,822 @@
...
+                    {
+                        "api_id": graphql_api.id,
+                        "domain_name": domain_name.domain_name,
+                    },
+                ),
+                opts=self._resource_opts(),
+            )
+
+            record = dns.create_record(
+                resource_name=prefix(f"{self.name}-domain-record"),
```
michal-stlv: safe_name,  limit is 255 chars.

README.md
```diff
@@ -72,6 +72,7 @@ Stelvio provides high-level components for the most common AWS services:
 - **[Message Queues](https://stelvio.dev/docs/guides/queues/)** (SQS)
 - **[Pub/Sub Topics](https://stelvio.dev/docs/guides/topics/)** (SNS)
 - **[Email](https://stelvio.dev/docs/guides/email/)** (SES)
+- **[AppSync](https://stelvio.dev/docs/guides/email/)** (AWS AppSync Managed GraphQL APIs)
```
michal-stlv: wrong link to email guide?

docs/guides/appsync.md
```diff
@@ -0,0 +1,842 @@
...
+
+### App-Level Customization Keys
+
+Use `StelvioAppConfig.customize` with per-component keys:
+
+- `AppSync` → `api`, `domain_name`, `api_key`
+- `AppSyncDataSource` → `data_source`, `service_role`
+- `AppSyncResolver` → `resolver`
+- `PipeFunction` → `function`
+```
```
michal-stlv: this shouldn't be here? it's not opening or closing any code block?

stelvio/aws/appsync/appsync.py
sebst: to_provider_config

stelvio/aws/appsync/appsync.py
sebst: to_provider_config

stelvio/aws/appsync/appsync.py
sebst: to_authorizer_config

stelvio/aws/appsync/appsync.py
sebst: fn_opts

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,720 @@
...
+    domain_dns_record: Record | None = None
+
+
+@final
+class AppSync(Component[AppSyncResources, AppSyncCustomizationDict], LinkableMixin):
+    def __init__(
+        self,
+        name: str,
+        config: AppSyncConfig | AppSyncConfigDict | None = None,
+        customize: AppSyncCustomizationDict | None = None,
```
michal-stlv: This should be keyword-only (after `*`) to match every other component in the codebase. Right now `customize` is positional while `tags` is keyword-only — should be  both keyword-only, tags first.

For reference, the `Api` constructor does it like this:
```python
def __init__(
    self,
    name: str,
    config: ApiConfig | None = None,
    *,
    tags: dict[str, str] | None = None,
    customize: ApiCustomizationDict | None = None,
    **opts: Unpack[ApiConfigDict],
) -> None:
```

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,720 @@
...
+        self._auth = parsed.auth
+        self._additional_auth = parsed.additional_auth
+        self._domain = parsed.domain
+
+        self._data_sources: dict[str, AppSyncDataSource] = {}
+        self._resolvers: list[AppSyncResolver] = []
+        self._resolver_keys: set[tuple[str, str]] = set()
+        self._pipe_functions: dict[str, PipeFunction] = {}
+        self._none_data_source: appsync.DataSource | None = None
+        self._auth_outputs: list[Output[Any]] = []
```
michal-stlv: `_auth_outputs` and `_domain_outputs` are only consumed by the test helper `when_appsync_ready` in `conftest.py`. Production code shouldn't carry test-only state.

The test helper already accesses data sources, resolvers, and pipe functions through their `.resources` — it can do the same for auth permissions and domain resources by reading from `api.resources.auth_permissions`, `api.resources.acm_validated_domain`, `api.resources.domain_association`, etc.

Please remove these two fields and update the test helper to derive what it needs from `api.resources` instead.

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,720 @@
...
+
+        auth_permissions = self._create_auth_permissions(
+            graphql_api,
+            auth_function,
+            additional_auth_functions,
+        )
+        self._auth_outputs = [permission.id for permission in auth_permissions]
+        api_key_resource = self._create_api_key(graphql_api)
+
+        self._none_data_source = appsync.DataSource(
```
michal-stlv: Setting `self._none_data_source` here is a side effect inside `_create_resources()`. The pattern across the codebase is that `_create_resources()` only creates resources and returns them — the base class handles assigning `self._resources`.

Would be cleaner to include `none_data_source` as a field on `AppSyncResources`:
```python
@dataclass(frozen=True)
class AppSyncResources:
    api: appsync.GraphQLApi
    api_key: appsync.ApiKey | None
    none_data_source: appsync.DataSource
    ...
```

stelvio/aws/appsync/data_source.py
```diff
@@ -0,0 +1,370 @@
...
+            customize=customize,
+        )
+
+        self._data_source_name = name
+        self._api = api
+        self._config = config
+
+    @property
+    def name(self) -> str:
+        return getattr(self, "_data_source_name", self._name)
```
michal-stlv: The `getattr` fallback to `self._name` can never trigger since `__init__` always sets `_data_source_name`. This pattern looks defensive against something that can't happen, which makes it harder to reason about.

Just `return self._data_source_name` would be clearer.

stelvio/aws/appsync/resolver.py
```diff
@@ -0,0 +1,202 @@
...
+    ) -> None:
+        super().__init__("stelvio:aws:PipeFunction", f"{api.name}-fn-{name}", customize=customize)
+        self._pipe_function_name = name
+        self._api = api
+        self._data_source = data_source
+        self._code = code
+
+    @property
+    def name(self) -> str:
+        return getattr(self, "_pipe_function_name", self._name)
```
michal-stlv: Same as the data source — `getattr` fallback can't trigger. `return self._pipe_function_name` is simpler and clearer.

stelvio/aws/appsync/data_source.py
```diff
@@ -0,0 +1,370 @@
...
+            function=function_instance,
+        )
+        self.register_outputs(
+            {
+                "name": self.name,
+                "arn": data_source.arn,
+                "service_role_arn": role.arn,
+            }
+        )
+        pulumi.export(
```
michal-stlv: Looking at the rest of the codebase, child/sub-components never call `pulumi.export()` — only the parent component does (e.g., `DynamoSubscription`, `QueueSubscription`, `TopicSubscription` have zero exports). The parent `AppSync` already exports `url`, `arn`, `id`, and `api_key` which is consistent with how other components work.

For an API with several data sources, resolvers, and pipe functions, these child exports would add a lot of noise to `stlv outputs`. I'd suggest removing the `pulumi.export` calls from `AppSyncDataSource`, `AppSyncResolver` (resolver.py:131), and `PipeFunction` (resolver.py:198).

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,720 @@
...
+    def _create_api_key(self, graphql_api: appsync.GraphQLApi) -> appsync.ApiKey | None:
+        api_key_auth = self._get_api_key_auth()
+        if api_key_auth is None:
+            return None
+
+        prefix = context().prefix
+        # Compute expiry from "now" so each deploy refreshes to a full validity window
+        # (bounded by ApiKeyAuth validation). This avoids near-expiry replacements during
+        # later updates and keeps rotation timing predictable.
+        expires_dt = datetime.now(tz=UTC) + timedelta(days=api_key_auth.expires)
```
michal-stlv, line 614: Just a heads-up (not blocking): since this recomputes on every deploy, `stlv diff` will always show a change on the API key resource even when nothing else changed. You could consider rounding to date-only granularity (`expires_dt.strftime("%Y-%m-%dT00:00:00Z")`) so deploys on the same day don't produce drift. But this is a known tradeoff — just wanted to flag it.

tests/integration/test_scenario_appsync.py
```diff
@@ -0,0 +1,168 @@
...
+}
+"""
+
+
+def _query_until_no_errors(
+    url: str,
+    query: str,
+    *,
+    api_key: str,
+    ready_check: callable,
```
michal-stlv: Lowercase `callable` is the builtin function, not a type annotation. Should be:
```python
from collections.abc import Callable
...
ready_check: Callable[[dict], bool],
```

pyproject.toml
```diff
@@ -102,6 +102,8 @@ known-first-party = ["stelvio"]
 [tool.ruff.lint.per-file-ignores]
 "tests/**/*.py" = [
     "S101", # asserts allowed in tests
+    "S106", # hardcoded password false positives in test data
```
michal-stlv: This suppresses hardcoded-password checks across *all* test files. Only a handful of spots actually trigger these (like the `next_token_arg="nextToken"` default in codegen.py). Prefer targeted `# noqa: S106` / `# noqa: S107` on the specific lines instead of blanket suppression — keeps the safety net for the rest of the test suite.

stelvio/aws/appsync/data_source.py
```diff
@@ -0,0 +1,370 @@
...
+@final
+@dataclass(frozen=True, kw_only=True)
+class AppSyncRdsSourceConfig:
+    cluster_arn: str
+    secret_arn: str
+    database: str
+
+
+@dataclass(frozen=True, kw_only=True)
+class AppSyncDataSourceTypeConfig:
```
michal-stlv, line 43: Not blocking, just a note for future consideration: this is a "bag" dataclass where only one of `handler`/`table`/`url`/`rds`/`endpoint` is populated based on `ds_type`. It means `_build_ds_type_config()` and `_attach_static_policies()` need runtime `if self._config.table is None: raise RuntimeError(...)` guards that wouldn't be necessary with separate config classes per data source type (e.g., `LambdaSourceConfig`, `DynamoSourceConfig`, etc.).

Fine for now since it's internal, but worth cleaning up if this module grows.

stelvio/aws/appsync/config.py
```diff
@@ -0,0 +1,267 @@
...
+    user_pool_id: str
+    region: str | None = None
+    app_id_client_regex: str | None = None
+
+    def __post_init__(self) -> None:
+        if not self.user_pool_id:
+            raise ValueError("user_pool_id cannot be empty")
+
+    def to_provider_config(self) -> dict[str, Any]:
+        config: dict[str, Any] = {"user_pool_id": self.user_pool_id}
```
michal-stlv, line 73: For default Cognito auth (GraphQLApi user_pool_config), Pulumi's input type requires default_action (ALLOW or DENY). This config currently omits it; can we set an explicit default (likely DENY) in CognitoAuth.to_provider_config() for the default-auth path?

docs/guides/appsync.md
```diff
@@ -0,0 +1,866 @@
...
+| `url`     | `Output[str]`     | GraphQL endpoint URL                              |
+| `arn`     | `Output[str]`     | API ARN                                           |
+| `api_id`  | `Output[str]`     | API ID                                            |
+| `api_key` | `Output[str] \| None` | API key value, or `None` if not configured    |
+
+## Runtime
+
+AppSync supports two resolver runtimes: APPSYNC_JS (JavaScript) and VTL (Velocity Template Language). Stelvio uses APPSYNC_JS exclusively — it's the modern standard and what AWS recommends for new APIs.
+
+If you need VTL for a specific resolver, use the customization escape hatch:
```
michal-stlv: This VTL fallback example looks inconsistent with current implementation: resolvers are created with APPSYNC_JS runtime/code by default, so setting only request_template/response_template here may not work as implied. Could we either document the exact supported escape hatch or remove this claim?

tests/integration/test_scenario_appsync.py
```diff
@@ -0,0 +1,168 @@
...
+            "errors" not in response and response.get("data", {}).get("listItems") is not None
+        ),
+    )
+    items = scan_result["data"]["listItems"]["items"]
+    assert len(items) >= 1
+    assert any(i["pk"] == "item-1" for i in items)
+
+
+def test_scenario_appsync_link_env_vars(stelvio_env, project_dir):
+    """Linking a Function to AppSync injects STLV_ env vars and IAM permissions."""
```
michal-stlv, line 153: The test title/docstring says we verify link env vars and IAM permissions, but the assertions only check env vars. Can we add an IAM assertion (for example via assert_lambda_role_permissions) or narrow the test description?

docs/guides/appsync.md
```diff
@@ -0,0 +1,866 @@
...
+api = AppSync("myapi", schema="schema.graphql",
+    auth="iam",
+    customize={
+        "api": {"xray_enabled": True},
+    },
+)
+```
+
+!!! info "Constructor scope"
+    Constructor-level `customize` applies only to top-level AppSync resources (`api`, `domain_name`, `api_key`).
```
michal-stlv: This sentence says constructor-level customize applies only to api/domain_name/api_key, but this page also documents auth_permissions/acm_validated_domain/domain_association/domain_dns_record as constructor-level keys. Could we make these sections consistent?

docs/guides/appsync.md
```diff
@@ -0,0 +1,866 @@
...
+
+```python
+from stelvio.aws.appsync import AppSync, CognitoAuth, ApiKeyAuth
+
+api = AppSync("myapi", schema="schema.graphql",
+    auth=CognitoAuth(user_pool_id="us-east-1_ABC123"),
+    additional_auth=["iam", ApiKeyAuth()],
+)
+
+posts = api.data_source_lambda("posts", handler="resolvers/posts.handler", links=[table])
```
michal-stlv: table is referenced in this quick-start (links=[table]) but not defined in the snippet. Could we either define it or remove it to keep the example copy-paste-safe?

tests/aws/appsync/test_appsync.py
```diff
@@ -0,0 +1,297 @@
...
+from stelvio.aws.appsync.constants import AUTH_TYPE_COGNITO
+from stelvio.config import AwsConfig
+from stelvio.context import AppContext, _ContextStore
+
+from .conftest import COGNITO_USER_POOL_ID, INLINE_SCHEMA, when_appsync_ready
+
+TP = "test-test-"
+
+
+@pulumi.runtime.test
```
michal-stlv, line 31: Can we add the standard config parity test here (same pattern as other components)?

assert_config_dict_matches_dataclass(AppSyncConfig, AppSyncConfigDict)

We already use this helper for Api/Queue/Dynamo/Function config pairs, and AppSync should get the same guard.

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,720 @@
...
+        name: str,
+        config: AppSyncConfig | AppSyncConfigDict | None = None,
+        customize: AppSyncCustomizationDict | None = None,
+        *,
+        tags: dict[str, str] | None = None,
+        **opts: Unpack[AppSyncConfigDict],
+    ) -> None:
+        super().__init__("stelvio:aws:AppSync", name, tags=tags, customize=customize)
+
+        parsed = self._parse_config(config, opts)
```
michal-stlv: Question on config pattern consistency: in other components with Config + ConfigDict pairs (Api, Queue, DynamoTable, Function, Email) we store parsed config as self._config and expose a read-only config property.

Here we decompose into separate fields directly. Any reason not to keep self._config + config property and derive only transformed schema text? This would reduce state surface and align with existing conventions.

stelvio/aws/appsync/appsync.py
```diff
@@ -0,0 +1,720 @@
...
+        if isinstance(config, AppSyncConfig):
+            return config
+        if isinstance(config, dict):
+            return AppSyncConfig(**config)
+
+        raise TypeError(
+            f"Invalid config type: expected AppSyncConfig or dict, got {type(config).__name__}"
+        )
+
+    def _check_not_created(self) -> None:
```
michal-stlv: AppSync should not freeze builder methods after parent resource creation (`_check_not_created` on `data_source_*`, `query/mutation/resolver`, `pipe_function`).

For `Api`/`Bucket` this freeze is valid because they compile a finalized builder graph into parent-owned resources.
AppSync is modeled differently: data sources/resolvers/pipe functions are independent child components that already resolve parent API resources lazily (`self._api.resources...`), matching the existing parent+child pattern used by `DynamoTable`/`Queue`/`Topic` subscriptions.

So enforcing freeze here is unnecessary and introduces avoidable special cases (for example around NONE data source lifecycle).

Direction for consistency:
- parent AppSync creates only parent resources
- child components remain addable and resolve parent lazily
- no additional “freeze” rule beyond normal component semantics

This keeps AppSync aligned with the framework’s existing component lifecycle model and avoids a custom behavior users need to remember.

michal-stlv (changes requested):
Thanks for the latest pass — this is much better and the DRY reduction is clear.

I still need one final cleanup pass before merge:

1. `docs/guides/appsync.md`
Please clarify the app-level customization note so it matches the real `StelvioAppConfig.customize` shape (component classes as top-level keys), and doesn’t read like string-key prefixing.

2. `tests/aws/appsync/test_appsync_data_sources.py`
Please restore explicit OpenSearch config-key coverage:
- `assert "opensearchserviceConfig" in ds_inputs`
- `assert ds_inputs["opensearchserviceConfig"]["endpoint"] == "https://search-my-domain-abc.us-east-1.es.amazonaws.com"`
- `assert "elasticsearchConfig" not in ds_inputs`

3. `tests/aws/appsync/test_appsync_data_sources.py`
Please add back a positive VPC-style OpenSearch endpoint test (for example `https://vpc-mydomain-...us-east-1.es.amazonaws.com`) and assert the derived IAM resource ARN exactly:
- `arn:aws:es:us-east-1:*:domain/mydomain/*`

4. `tests/aws/appsync/test_appsync_codegen.py`
Please tighten at least the baseline scan/query assertions back to exact-template checks (or normalized exact checks), not fragment-only checks.

After these are in, I’m happy to re-review quickly.

PR #164, opened 2026-03-03
feat: add Cognito components

sebst:


michal-stlv (changes requested):
Hey Bas — solid work, here's the detailed pass.

## Follow-up after CLI branch merges

These depend on `feat/cli-output-redesign` (not on main yet), handle after that lands:

**1. Trigger Functions missing `parent=self`** — `_create_trigger_function` creates Functions without `parent=self`. Other components (Cron, subscriptions) pass it.

**2. Missing `RESOURCE_TYPE_NAMES` entries** — `rich_deployment_handler.py` needs entries for `aws:cognito/userPool:UserPool`, `aws:cognito/userPoolClient:UserPoolClient`, `aws:cognito/identityProvider:IdentityProvider`.

**3. Missing `_DATA_LOSS_REPLACEMENT_TYPES` entry** — UserPool contains user data. `aws:cognito/userPool:UserPool` should be in the data loss warning set.

## Must fix

**4. `IdentityProvider` missing from keyword-only constructor params test** — add `(IdentityProvider.__init__, "customize")` to `test_keyword_only_constructor_params.py`.

**5. Docs nav not updated** — `zensical.toml` doesn't include `docs/components/aws/cognito.md`.

**6. Changelog link is wrong** — points to `guides/cognito.md`, should be `components/aws/cognito.md`.

## Should fix

**7. `autoVerifiedAttributes` tests use `in` instead of exact equality** — 4 tests in `test_user_pool.py` use `assert "email" in mock.inputs["autoVerifiedAttributes"]` which passes with unexpected extra entries. Use exact list equality.

**8. Trigger tests missing `Output.all()` synchronization** — several tests in `test_user_pool_triggers.py` wait only on `pool.arn` but check trigger functions/permissions. Include all relevant outputs in `Output.all()`.

**9. `_pool_resource` cross-component mutation** — `_prepare_children` sets private attrs on children with `noqa: SLF001`. Add a docstring on `_pool_resource` in UserPoolClient and IdentityProvider explaining the intentional coupling.

**10. `Email` import not TYPE_CHECKING-guarded in `types.py`** — `from stelvio.aws.email import Email` is only used as a type annotation, should be behind `TYPE_CHECKING`.

**11. Unused `POOL_ARN_TEMPLATE` in `test_user_pool.py`** — defined but never used (it's used in `test_links.py`). Remove.

**12. `UserPool` missing from `SIMPLE_COMPONENTS` in `test_type_urns.py`**.

**13. `IdentityProviderConfig` missing `kw_only=True`** — inconsistent with other config dataclasses.

## Consider

**14.** Link permissions are read-only (`GetUser`, `AdminGetUser`, `ListUsers`) — many real-world Lambdas need admin actions. Docs note this which is good, just flagging.

**15.** `test_fully_loaded_config` checks 7 of 9 configured properties — doesn't verify `autoVerifiedAttributes` or `softwareTokenMfaConfiguration`.

**16.** No integration test for SES email config (likely deferred due to SES verification requirements).

**17.** Forward-looking mock code for IdentityPool in `pulumi_mocks.py` is unused — fine as Phase 2 prep.

**18.** Manual type annotation comparison in `test_user_pool_config_dict_matches_dataclass` uses fragile string representation — add a comment noting Python 3.12+ specificity.

michal-stlv (commented):
One more thing — the Sign-in Options docs section could use a bit more context for readers who are new to Cognito:

- **`usernames` mode**: worth clarifying that there is no username at all — Cognito generates a random UUID internally and the email/phone IS what users type to log in. Right now "no separate username field" is easy to miss.
- **`aliases` mode**: explain that the user picks a username during sign-up (like `john42`) and it's fixed forever. The aliases are just additional ways to sign in. Currently the reader has to guess how the username gets set.
- **`preferred_username`**: this one appears without explanation. It's a Cognito-specific concept — a changeable display name the user can sign in with (unlike the fixed internal username). A sentence explaining what it is and when you'd use it would help.
- **Why mutually exclusive**: one sentence on the fundamental difference — "email IS your identity" vs "email is an alternative to your username" — would make the choice much clearer.

The info box recommending `usernames=["email"]` for most apps is good — just needs more context around the two modes so beginners can understand the trade-off.

michal-stlv (dismissed):

michal-stlv (changes requested):
Final review before merge to main — almost there, few things to fix:

**Must fix:**

1. **`pulumi.export` calls still in user_pool.py and user_pool_client.py** — removed from identity_pool.py but missed in the other two. user_pool.py exports id+arn, user_pool_client.py exports id+user_pool_id.

2. **Broken relative links in cognito.md** — `linking.md` and `customization.md` should be `../../concepts/linking.md` and `../../concepts/customization.md` (4 occurrences). Other component docs in the same directory use the correct paths.

3. **Changelog should mention IdentityPool** — currently only says "Cognito User Pools" but IdentityPool is a distinct component in this PR.

**Should fix:**

4. **Nested `.apply()` in link permission tests** — `test_user_pool_link_permissions` and `test_client_link_permissions` do `permission.resources[0].apply(verify_resource)` inside an already-running `.apply()` callback. Inner assertion failures can be silently swallowed. Resolve via `Output.all()` at the top level.

5. **Missing config dict parity tests for IdentityPool** — `IdentityPoolConfig`/`IdentityPoolConfigDict` and `IdentityPoolPermissions`/`IdentityPoolPermissionsDict` don't have the parity tests that all other config types have.

**Minor:**

6. `test_canonical_list_has_26_entries` name says 26 but asserts 27.

7. Changelog date placeholder `(2026-04-DD)` needs filling in as well as version number etc.

michal-stlv (approved):

PR #166, opened 2026-03-04
feat: tagging support with global and per-component tags

michal-stlv:


stelvio/component.py
```diff
@@ -44,11 +46,29 @@ def __init__(
...
+        normalized: dict[str, str] = {}
+        for key, value in tags.items():
+            if not isinstance(key, str):
+                raise TypeError(f"Tag key must be str, got {type(key).__name__} in tags={tags!r}")
+            if not isinstance(value, str):
+                raise TypeError(
+                    f"Tag value for key '{key}' must be str, got {type(value).__name__}"
+                )
+            normalized[key] = value
+        return normalized
```
sebst: could just return tags

tests/test_cleanup_aws.py
```diff
@@ -7,8 +7,8 @@
 import sys
 from pathlib import Path
 
-# cleanup_aws is a standalone script, not a package module — add its directory
-sys.path.insert(0, str(Path(__file__).resolve().parent))
+# cleanup_aws is a standalone script in tests/integration/, not a package module
+sys.path.insert(0, str(Path(__file__).resolve().parent / "integration"))
```
sebst: why move it? the cleanup is in the scope of integration tests, no?

stelvio/component.py
```diff
@@ -44,11 +46,29 @@ def __init__(
...
         ComponentRegistry.add_instance(self)
 
+    @staticmethod
+    def _normalize_tags(tags: dict[str, str] | None) -> dict[str, str]:
+        if tags is None:
+            return {}
+        if not isinstance(tags, dict):
+            raise TypeError(f"tags must be a dict[str, str], got {type(tags).__name__}")
+        normalized: dict[str, str] = {}
+        for key, value in tags.items():
```
sebst: user is able to overwrite auto-tags (stelvio:*). intentional?

stelvio/component.py
michal-stlv: Good catch — dropped the redundant dict rebuild. Renamed to `_validate_tags` since it's validation, not normalization. It validates in-place now and doesn't return anything.

tests/test_cleanup_aws.py
michal-stlv: Removed the sys.path hack entirely — `tests/integration/` has `__init__.py` so a normal package import works fine. This test file lives in `tests/` (not `tests/integration/`) because it's a unit test for the cleanup script's logic — it doesn't need AWS credentials or real resources, just tests ARN parsing, name matching, etc.

stelvio/component.py
michal-stlv: Yes, intentional. Precedence is component > global > auto-tags. It's the user's infrastructure — if they deliberately override a `stelvio:*` tag, we respect that. Happy to hear if you think differently though.

PR #187, opened 2026-04-09
chore: ensure compat with python versions

sebst:
Main issue: Python 3.14 removed implicit event loop creation from `get_event_loop()`.

Pulumi's SDK internally calls `asyncio.get_event_loop()`, which now raises `RuntimeError` when no current loop is set.

We introduce an event loop fixture for our tests, which has also uncovered flaky tests in Python 3.14 runtimes.

Also added matrix to test workflow to cover Python versions 3.12, 3.13, 3.14 on Ubuntu (Intel and ARM).


stelvio/aws/api_gateway/api.py
```diff
@@ -59,6 +59,9 @@ class ApiResources:
     rest_api: RestApi
     deployment: Deployment
     stage: Stage
+    custom_domain: "DomainName | None" = None
+    base_path_mapping: "BasePathMapping | None" = None
+    permissions: "list[Permission]" = field(default_factory=list)
```
michal-stlv: are these here only because of tests? is thise useful to user ? do these still map to customize and can be changed with it?

stelvio/aws/appsync/data_source.py
```diff
@@ -29,6 +29,7 @@ class AppSyncDataSourceResources:
     data_source: "appsync.DataSource"
     service_role: "iam.Role"
     function: "Function | None" = None
+    policies: "list[iam.RolePolicy]" = field(default_factory=list)
```
michal-stlv: same here, only for tests? useful for users? part of customize etc?

michal-stlv:
- In stelvio/aws/api_gateway/api.py, the PR starts tracking API
    Lambda Permission resources in ApiResources.permissions, but from
    the diff that list is only populated for route-handler permissions
    created in get_group_function().
- In tests/aws/api_gateway/conftest.py, the new when_api_ready()
    helper waits on api.resources.permissions.
- In tests/aws/api_gateway/test_api_authorizers.py, authorizer tests
    are switched to that helper too, but authorizer permissions come
    from _create_authorizer_permission() and do not appear to be added
    to the tracked list.

  That means the PR’s “wait for all async Pulumi side resources before
  asserting” fix looks incomplete for authorizer-permission tests. It
  may still pass CI, but it leaves exactly the kind of flake this PR is
  trying to remove.

michal-stlv (approved):

PR #189, opened 2026-04-10
feat: improve cognito

sebst:


michal-stlv (dismissed):

michal-stlv (commented):
Few things to consider:

### 1. StelvioTestEnv double-deploy guard removed

The old guard in `stelvio_test_env.py` prevented accidental double-deploy. Removing it for `test_identity_pool_string_binding` is fine, but future test authors might double-deploy accidentally, or a failure between deploys could orphan resources. Consider adding a `redeploy()` method to make the intent explicit, or at minimum a docstring explaining the semantics.

### 2. CognitoAuth validation gap

`__post_init__` in `appsync/config.py` only validates empty strings. Passing `None` or an integer silently passes and fails later in `to_provider_config()` with an opaque error. An `else` branch raising on unexpected types would catch mistakes at construction time:

```python
def __post_init__(self) -> None:
    from stelvio.aws.cognito.user_pool import UserPool
    if isinstance(self.user_pool_id, str):
        if not self.user_pool_id:
            raise ValueError("user_pool_id cannot be empty")
    elif not isinstance(self.user_pool_id, UserPool):
        raise TypeError(
            f"user_pool_id must be a UserPool or string, got {type(self.user_pool_id).__name__}"
        )
```

### 3. test_identity_pool_string_binding doesn't verify providerName

The integration test asserts `expected_provider_count=1` but never checks the actual `providerName` contains the correct parsed region. The core behavior (`pool_id.split("_")[0]` for region) is not validated end-to-end — the test would pass even if region parsing were wrong.

### 4. assert_appsync_api has_user_pool_config=False silently passes

Uses `if has_user_pool_config:` instead of `if has_user_pool_config is not None:`. Passing `False` to assert there's NO pool config does nothing. Should handle the negative case:

```python
if has_user_pool_config is not None:
    pool_config = api.get("userPoolConfig", {})
    pool_id = pool_config.get("userPoolId", "")
    if has_user_pool_config:
        assert pool_id, "Expected userPoolConfig with non-empty userPoolId"
    else:
        assert not pool_id, f"Expected no userPoolConfig, got userPoolId={pool_id}"
```

### 5. IdentityPoolBinding underscore validation is permissive

The `"_" not in self.user_pool` check catches the obvious case but accepts `_foo` or `foo_` where `split("_")[0]` produces empty/invalid region. Fine for now — catches the common mistake.

michal-stlv (approved):

PR #191, opened 2026-04-10
feat: make app.config optional

sebst:


michalmartinka (dismissed):
.

michal-stlv (changes requested):
**1. Initialize `self._app_config` in `__init__`**
Declare `self._app_config: StelvioAppConfig | None = None` in `__init__` and drop the inline annotation from the existing assignment. Without this, accessing `self._app_config` before `_execute_user_config_func` runs raises `AttributeError` instead of a clear `None` — and having the annotation split across two branches is inconsistent.

**2. Add `home` field assertion to default config test**
`test_default_config_when_no_app_config_registered` checks every field except `home`. This field defaults to `"aws"` and drives which storage backend is used in `command_run.py` — if the default ever changes accidentally, this test should catch it. Add `assert config.home == "aws"`.

**3. Add a test for `env` passthrough**
`test_user_config_returned_when_app_config_registered` never verifies that `env` actually reaches the user's callback — it ignores the argument and returns a fixed object. The whole point of `@app.config` receiving `env` is per-environment config (different profiles for staging vs prod). If a future change accidentally dropped `env` forwarding, this gap would miss it. Add a test that captures the received env and asserts it matches what was passed.

**4. Document shared-environments constraint**
With no `@app.config`, `environments=[]` means only the user's personal env (their username) is valid. Running `stlv deploy prod` without `@app.config` gives a confusing error referencing "configured shared environments" that were never configured. Add a note in the docs tip that shared/named environments require `@app.config`.

michal-stlv (approved):

PR #202, opened 2026-05-27
docs: rewrote stelvio-app.md to make sense and  fixes in other places

michal-stlv:


docs/docs/concepts/stelvio-app.md
```diff
@@ -1,252 +1,88 @@
...
 
-**Use environment variables (CI/CD)**:
-```python
-# Set in environment: AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_REGION
-@app.config
-def configuration(env: str) -> StelvioAppConfig:
-    return StelvioAppConfig()  # Uses env vars automatically
-```
+!!! warning "Environment variables must be exported"
+    Just setting `AWS_PROFILE=my_profile` is not enough. It only sets a shell variable. Subprocesses like `stlv` will not see it. You need to use `export AWS_PROFILE=my_profile`. Verify by running `env | grep AWS`.
```
sebst: could be a good oppurtunity to change the CLI in the docs from `stlv` to `stelvio`?

sebst: i find this comment a bit confusing... 
you're talking explicitly about this case, i assume:

```bash
AWS_PROFILE=my_profile
stlv
...
```

However, apparently
```bash
AWS_PROFILE=my_profile stlv
```
would work just fine.

"Subprocesses like `stlv`" is the part that confused me, also "Environment variables must be exported", while technically, exporting a variable is what makes it an environment variable.

Maybe we could change from "warning" to "tip":

> Check your environment variables by running `env | grep AWS`. Use `export AWS_PROFILE=my_profile` to set your environment variables

michal-stlv: I agree with change but I think it should be separate PR just with rename. To keep PRs managable.

michal-stlv: yes that's what I talk about but when i do just 
```bash
AWS_PROFILE=my_profile stlv

uv run stlv deploy
```

that does NOT work for me. I have to do `export AWS_PRfOFILE=my_profile`.

Have you tried ? Does simple `AWS_PROFILE=my_profile` work for you?

I'm happy to update wording. Do you have any suggestion? But I really need `export` otherwise it does not work. I'm on latest macos and latest stable ghostty.

sebst (dismissed):
left two nps

docs/docs/concepts/stelvio-app.md
sebst: > Does simple AWS_PROFILE=my_profile work for you?

that's what I meant: two-line does not work; one line (`AWS_PROFILE=my_profile stlv deploy`) works.

`export` is what "promotes" a variable to an env var.

suggestion was:

Maybe we could change from "warning" to "tip":

Check your environment variables by running env | grep AWS. Use export AWS_PROFILE=my_profile to set your environment variables

michal-stlv: I kept it warning but updated title and content and made it collapsed. Ok?

sebst (approved):

PR #204, opened 2026-06-05
feat: vpc component

michal-stlv:
First piece of VPC support: a `Vpc` component that creates a `10.0.0.0/16` VPC with public, private and isolated subnets across AZs, a route table per subnet, and optional managed NAT — one per AZ, a single shared one, or with your own Elastic IPs. Docs and changelog included. Lambda-in-VPC, DocumentDB and automatic security groups via linking come in follow-up PRs.

While writing tests I got tired of our `created_*` mock helpers, so pulumi mocks got a small new assertion API: `assert_res` / `assert_no_res` / `assert_res_counts` plus an `R` resource-type enum. One call asserts a resource and its inputs, and counts are sealed so new resources can't sneak in unnoticed. I prefer this enough that I'd like to gradually migrate existing tests to it and remove the ~60 `created_*` methods.


stelvio/aws/vpc.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+SUBNET_LAYOUTS: Final[dict[SubnetType, SubnetLayout]] = {
+    SubnetType.PUBLIC: SubnetLayout(24, 0),
+    SubnetType.PRIVATE: SubnetLayout(22, 20),
+    SubnetType.ISOLATED: SubnetLayout(24, 60),
+}
+
+
+@dataclass(frozen=True)
```
sebst: np: `kw_only=True`

stelvio/aws/vpc.py
```diff
@@ -0,0 +1,407 @@
...
+                elastic_ips.append(eip)
+                eip_allocation_id = eip.allocation_id
+            nat = self._create_nat_gateway(igw, az, eip_allocation_id, public_subnet)
+            nat_gateways.append(nat)
+
+        # one default route per private route table; single NAT → all share it
+        for i, private_rt in enumerate(route_tables_dict[SubnetType.PRIVATE]):
+            nat = nat_gateways[0] if nat_config.single else nat_gateways[i]
+            Route(
+                context().prefix(f"{self.name}-nat-route-{azs[i][-1]}"),
```
sebst, line 297: `safe_name` not needed?

docs/docs/components/aws/vpc.md
```diff
@@ -0,0 +1,273 @@
...
+| `elastic_ip`            | [EipArgs](https://www.pulumi.com/registry/packages/aws/api-docs/ec2/eip/#inputs)                                    | NAT Elastic IPs                 |
+| `nat_gateway`           | [NatGatewayArgs](https://www.pulumi.com/registry/packages/aws/api-docs/ec2/natgateway/#inputs)                      | NAT Gateways                    |
+
+### Example
+
+```python
+vpc = Vpc(
+    "main",
+    customize={
+        "public_subnet": {"map_public_ip_on_launch": True},
```
sebst, line 175: this would be the only way to trigger provision of a public ip, no? shouldn't that be the default case for `SubnetType.PUBLIC`? maybe promote it to config level argument?

stelvio/aws/vpc.py
```diff
@@ -0,0 +1,407 @@
...
+                nat_gateway_id=nat.id,
+                opts=self._resource_opts(),
+            )
+
+        return elastic_ips, nat_gateways
+
+    def _create_nat_gateway(
+        self, igw: InternetGateway, az: str, eip_allocation_id: Input[str], public_subnet: Subnet
+    ) -> NatGateway:
+        nat_name = self._safe_name(f"-nat-{az[-1]}")
```
sebst, line 309: wouldn't that create collisions for AZs like `us-east-1-iah-1a` and `us-east-1-atl-1a`?

stelvio/aws/vpc.py
```diff
@@ -0,0 +1,407 @@
...
+        super().__init__("stelvio:aws:Vpc", name, tags=tags, customize=customize)
+        _validate_az(az)
+        self._az = az
+        self._nat_config = _normalize_nat(nat)
+        _validate_nat_config(self._nat_config, self._az)
+
+    def _create_resources(self) -> VpcResources:
+        vpc = self._create_vpc()
+        igw = self._create_internet_gateway(vpc)
+        azs = _get_az_names(self._az)
```
sebst: `_get_az_names` potentially raises `ValueError`, but at that time, `vpc`, and `igw` are already created

tests/aws/pulumi_mocks.py
```diff
@@ -258,6 +347,11 @@ def call(self, args: MockCallArgs) -> tuple[dict, list[tuple[str, str]] | None]:
                 "region": "us-east-1",
                 "description": "US East (N. Virginia)",
             }, []
+        if args.token == "aws:index/getAvailabilityZones:getAvailabilityZones":  # noqa: S105
+            return {"names": ["us-east-1a", "us-east-1b", "us-east-1c"]}, []
+
+        if args.token == "aws:index/getRegion:getRegion":  # noqa: S105
```
sebst: duplicate (line 344)

.python-version
```diff
@@ -0,0 +1 @@
+3.12
```
sebst, line 1: should not be in the repo, should it?

pyproject.toml
```diff
@@ -91,7 +91,7 @@ select = ["E", "F", "I", "B", "N", "UP", "C4", "A", "S", "ARG", "LOG", "G", "PIE
     "PGH", "PL", "FURB", "RUF", "TRY", "INP", "ANN"
     # "D" - will need, defo need to work on more docs
 ]
-ignore = ['TRY003']
+ignore = ['TRY003', 'PT013']
```
sebst, line 94: should be a refactoring, no?

zensical.toml
```diff
@@ -113,6 +113,7 @@ Components = [
         { "Email" = "docs/components/aws/email.md" },
         { "Cognito" = "docs/components/aws/cognito.md" },
         { "AppSync" = "docs/components/aws/appsync.md" },
+        { "Vpc" = "docs/components/aws/vpc.md" },
```
sebst, line 116: missing entry in README.md

sebst (changes requested):

stelvio/aws/vpc.py
michal-stlv: good catch, fixed

stelvio/aws/vpc.py
michal-stlv: not needed. it's correct as it is.route does not have name in aws and also we do not add tags so no tag length limit either.  only reason to change would be consistency with other resoruces in vpc and maybe less confusing on the first look. but then maybe more confusion when you dig deeper into what needs safe name and what not.

docs/docs/components/aws/vpc.md
michal-stlv: thought about it - it’s just a launch default for ec2 instances, and nothing stelvio puts in public subnets uses it (nat brings its own EIP, lambda never gets a public ip). even later with ec2/containers the standard setup is alb in public + compute in private, so the flag likely stays unused. and if you do launch an instance you can set the ip on the instance itself, which is more explicit than a subnet-wide default. so keeping it off - customize covers the rare case (that’s the docs example).

stelvio/aws/vpc.py
michal-stlv: good catch, this are are local zones. they’d collide, but local zones don’t work with this component anyway (can’t host NAT, we create one per AZ). Basically, we do not support them. if local zones become a real ask it’s a proper feature (NAT stays in region AZs etc), not a naming fix. might add a docs note that az expects standard AZs but it's so niche that I don't even want to do that :)

stelvio/aws/vpc.py
michal-stlv: good catch, fixed - moved the az lookup before vpc/igw creation so it fails before anything is created

tests/aws/pulumi_mocks.py
michal-stlv: good catch, fixed

.python-version
michal-stlv: I did this intentionally, uv reads it so local dev always runs 3.12, the lowest supported version. that way you can’t accidentally use 3.13+ stuff that’d break users on 3.12. CI still tests the full 3.12-3.15 matrix. committing it is the uv recommended setup

pyproject.toml
michal-stlv: Not in my opinion, I think that rule is stupid.  I prefer from pytest import param, raises, mark, shorter call sites and parametrize tables read way better than pytest.param(...) everywhere. PT013 is purely stylistic, no functional difference. airflow ignores it too with the comment “silly rule prohibiting e.g. from pytest import param” :) hypothesis ignores it as well. older files here still use import pytest, they can migrate over time

zensical.toml
michal-stlv: good catch . also added it for Router

sebst (approved):

PR #208, opened 2026-06-12
feat: support customize with callable

### LINKED ISSUE #205: support customize with callable
We want to support customization witch callable (function, lambda) in addition to dict.

As agreed it will take dict of values that Stelvio is gonna pass to pulumi resource and user has a chance to modify them.

Later we might want to consider use of Pulumi's `*Args` objects rather than dict as param to customize callable.

sebst:


stelvio/component.py
```diff
@@ -202,12 +203,20 @@ def _customizer(
...
+        final_props = dict(default_props)
+        if global_customize := context().customize.get(type(self)):
+            if isinstance(global_customize, Callable):
+                final_props = _normalize(global_customize(default_props))
+            else:
+                final_props |= _normalize(global_customize.get(resource_name))
+
+        if local_customize := self._customize.get(resource_name):
+            if isinstance(local_customize, Callable):
+                final_props = _normalize(local_customize(default_props))
```
michal-stlv: Shoudn't we pass `final_props` here rather than `default_props`? It's multi level customization, local customizer should see values that Stelvio is gonna use which at this point are those modified by global customization.
Also worth having test for that.

michal-stlv (changes requested):
I think we should be more precise than Callable. We know params, and output type.

stelvio/component.py
```diff
@@ -205,12 +206,20 @@ def _customizer(
...
-            **default_props,
-            **_normalize(global_customize.get(resource_name)),
-            **_normalize(self._customize.get(resource_name)),
-        }
+        final_props = dict(default_props)
+        if global_customize := context().customize.get(type(self)):
+            if isinstance(global_customize, Callable):
+                final_props = _normalize(global_customize(final_props))
+            else:
+                final_props |= _normalize(global_customize.get(resource_name))
```
michal-stlv: This is wrong. 
This code expects that `global_customize` might be Callable but it should never be. As as also documented. It must be always dict. Only within the dict there might be Callable or dict for specific key. Same as for local. so this needs to be fixed. And also tests updated/added.

```suggestion
        global_component_customize = context().customize.get(type(self))
        if global_customize := global_component_customize.get(resource_name):
            if isinstance(global_customize, Callable):
                final_props = _normalize(global_customize(final_props))
            else:
                final_props |= _normalize(global_customize)
```

`test_customizer_applies_global_callable_customization` and `test_customizer_global_and_local_callables_are_both_invoked` are bad tests then.

stelvio/aws/email.py
```diff
@@ -44,11 +45,36 @@ class EmailResources:
...
+        pulumi_aws.ses.DomainIdentityVerificationArgs
+        | dict[str, Any]
+        | Callable[
+            [dict[str, Any]],
+            dict[str, Any] | pulumi_aws.ses.DomainIdentityVerificationArgs,
+        ]
+        | None
+    )
     event_destinations: (
         pulumi_aws.sesv2.ConfigurationSetEventDestinationArgs | dict[str, Any] | None
```
michal-stlv: this is missing Callable form

stelvio/aws/api_gateway/api.py
```diff
@@ -60,11 +60,36 @@ class ApiResources:
 
 
 class ApiCustomizationDict(TypedDict, total=False):
-    rest_api: pulumi_aws.apigateway.RestApiArgs | dict[str, Any] | None
-    deployment: pulumi_aws.apigateway.DeploymentArgs | dict[str, Any] | None
-    stage: pulumi_aws.apigateway.StageArgs | dict[str, Any] | None
-    custom_domain: pulumi_aws.apigateway.DomainNameArgs | dict[str, Any] | None
-    base_path_mapping: pulumi_aws.apigateway.BasePathMappingArgs | dict[str, Any] | None
+    rest_api: (
+        pulumi_aws.apigateway.RestApiArgs
```
michal-stlv: Why not just import `RestApiArgs` ? If there's a good reason not to import only types we need (what that would be?) Then import closest module and if it's long then alias it to something shorter. This is pain to read and takes lot of space. 
Applies to all the places that do this in this PR. And unfortunately in other places that our out of scope for this PR. Slipped in previous PRs reviews.
Let's fix it in places that this PR touches - type annotations of customization dicts.

stelvio/aws/api_gateway/api.py
```diff
@@ -60,11 +60,36 @@ class ApiResources:
...
+        | dict[str, Any]
+        | Callable[[dict[str, Any]], dict[str, Any] | pulumi_aws.apigateway.RestApiArgs]
+        | None
```
michal-stlv: Repeating this long type annotation tens of times is PITA. Long, pain to type, pain to read, pain to copy paste even. Let's use this:
```python
type Customization[T] = T | dict[str, Any] | Callable[[dict[str, Any]], T | dict[str,
  Any]] | None
```
Then we can use it as:
```python
...
rest_api: Customization[RestApiArgs]
...
```

For keys that don't have `*Args` we can still type it or have nother type e.g. `CustomizationNoArgs`

michal-stlv (changes requested):
- Global customization done wrong. 
- Type annotations should use generic type alias `Customization[T]`
- We should import what we need not whole module
- Type annotations not updated in for `customize` param in `StelvioAppConfig`. 
- docs - I don't remember if we discussed it'd be separate PR?
- same for changelog

stelvio/component.py
sebst: fixed

stelvio/aws/email.py
sebst: fixed

stelvio/aws/api_gateway/api.py
sebst: > Let's fix it in places that this PR touches - type annotations of customization dicts. 

last commit added `if TYPE_CHECKING` to the components touched and improved other places, too, along the way (`from __future__ import annotations`)

stelvio/aws/api_gateway/api.py
sebst: Put that in a separate module to better handle circular imports (moved out of `component.py`)

michal-stlv: Why was this problem with circular imports? Aren't `*CustomizationDict` classes next to component classes that inherits from Component in component.py ? So no new imports, or what am I missing?

sebst: > Why was this problem with circular imports? Aren't `*CustomizationDict` classes next to component classes that inherits from Component in component.py ? So no new imports, or what am I missing?

`stelvio.context` imports `StelvioAppConfig` from `stelvio.config`
`stelvio.config` imports `Customization` from `stelvio.component`
`stelvio.component` imports from `stelvio.context`

exact place is here: https://github.com/stelviodev/stelvio/pull/208/changes/dc90e4a09d10cff51a2f192583fa905204bf8c15#diff-b811ae9720d71cfd02257c9de0b602701c17e534e210c6508ef20bf57c7ab054L211

michal-stlv: ok, thanks. let's leave it in separate file then.

stelvio/component.py
```diff
@@ -205,12 +206,21 @@ def _customizer(
...
-        global_customize = context().customize.get(type(self), {})
-        return {
-            **default_props,
-            **_normalize(global_customize.get(resource_name)),
-            **_normalize(self._customize.get(resource_name)),
-        }
+        final_props = dict(default_props)
+        global_component_customize = context().customize.get(type(self), {})
+        if global_customize := global_component_customize.get(resource_name):
+            if isinstance(global_customize, Callable):
```
michal-stlv: we should use build in function`callable()`: `if callable(global_customize)` which seems to be idiomatic way to test if something is callable(also shorter, nicer). same below and everywhere. 

Then we can also move `from collections.abc import Callable` back under if TYPE_CHECKING

stelvio/component.py
michal-stlv: This should be updated to:
- tell user we suppor callable
- explain behaviour change between dict and callable: dict is shallow merge, callable is whole replace so user needs to return whole thing that will be used
- explain local cutomizer callable recieves props after global customization

Same thing needs to be explain in update docs, but since this is with code it's in the scope of this PR?

stelvio/aws/acm.py
```diff
@@ -18,11 +25,11 @@ class AcmValidatedDomainResources:
...
 
 class AcmValidatedDomainCustomizationDict(TypedDict, total=False):
-    certificate: pulumi_aws.acm.CertificateArgs | dict[str, Any] | None
+    certificate: Customization[CertificateArgs]
     validation_record: (
-        dict[str, Any] | None
-    )  # No specific Plumi Args type here, because cross cloud compat
-    cert_validation: pulumi_aws.acm.CertificateValidationArgs | dict[str, Any] | None
+        CustomizationNoArgs  # No specific Pulumi Args type here, because cross cloud compat
+    )
```
michal-stlv: Nitpick but can format this differently? 3 lines and brackets because of comment is just ugly imo.
```python
    # No specific Pulumi Args type here, because cross cloud compat
    validation_record: CustomizationNoArgs
```

michal-stlv (changes requested):
- context.py line 22 needs update for callable
- missing test for case where callable returns args
- missing test if callable returns `{}` - we have that test for dict
- negative test for global callable to customize only what it should - we have it for dict : `returns_default_props_when_resource_not_in_customize`

stelvio/component.py
sebst: fixed

stelvio/component.py
sebst: fixed

stelvio/aws/acm.py
sebst: fixed

michal-stlv (dismissed):
perfecto! gracias!

michal-stlv (dismissed):

michal-stlv (approved):

PR #211, opened 2026-06-20
fix: component.py type annotations

michal-stlv:


stelvio/component.py
```diff
@@ -43,11 +44,11 @@ def __init__(
         Args:
             type_name: Pulumi type URN (e.g., ``"stelvio:aws:Function"``).
             name: Globally unique component name.
-            tags: AWS tags applied to taggable child resources.
+            tags: AWS tags are applied to taggable child resources.
```
sebst: np: would leave it as is

sebst (dismissed):

stelvio/component.py
michal-stlv: hmm ok reads better without "are"

sebst (approved):

PR #212, opened 2026-06-20
feat: global customize to work as global defaults

sebst:


stelvio/aws/function/function.py
michal-stlv: This is confusing now as default_props are passed to computed_props param

stelvio/component.py
```diff
@@ -190,44 +201,65 @@ def _customizer(
...
-        final_props = dict(default_props)
         global_component_customize = context().customize.get(type(self), {})
-        if global_customize := global_component_customize.get(resource_name):
+        global_customize = global_component_customize.get(resource_name)
+        local_customize = self._customize.get(resource_name)
+
+        if default_props is None:
+            default_props = {}
+
+        # Defaults mode used by Function to treat global customize as defaults.
```
michal-stlv: I don't understand this comment. Function? This is not function specific code, is it?

stelvio/component.py
```diff
@@ -190,44 +201,65 @@ def _customizer(
...
+        final_props = effective_defaults | {
+            k: v for k, v in computed_props.items() if v is not None
+        }
+        if local_customize:
             if callable(local_customize):
-                final_props = _normalize(local_customize(final_props))
+                local_customize_fn = cast("Callable[[dict[str, Any]], object]", local_customize)
+                final_props = _normalize(local_customize_fn(final_props))
             else:
                 final_props |= _normalize(local_customize)
```
michal-stlv, line 264: Hmm so global callable does NOT recieve computed props, right? That's changed from #208 .

So the way I understand this:
1. we take default props and pass them to global customize (if callable)or merge with global customize (if dict)
2. we then merge  resulting dict and merge it with computed_props where we ignore computed props with value None, also computed props win over defaults if not None
3. then we take resulting dict and pass it to local customize (if callable) or merge with local customize (if dict)

Is that correct?

If yes, then:

- it's OK for case when global customize is dict. 
- BUT if global customize is callable I'd argue:
    - user can't extent some props globally - e.g. adding some envar to all functions
    - user actually don't have access to all props that component/resource is gonna use so can't make decision based on that: e..g set memory to X for all functions where handler starts with `src/processing`. Anoher case might be to add function layers for specific paths, etc. etc. 

I'm not sure that's desired behavior.

stelvio/component.py
```diff
@@ -190,44 +201,65 @@ def _customizer(
...
+                effective_defaults |= _normalize(global_customize)
 
-        if local_customize := self._customize.get(resource_name):
+        final_props = effective_defaults | {
+            k: v for k, v in computed_props.items() if v is not None
+        }
+        if local_customize:
             if callable(local_customize):
-                final_props = _normalize(local_customize(final_props))
+                local_customize_fn = cast("Callable[[dict[str, Any]], object]", local_customize)
```
michal-stlv: this type cast is not needed is it? i dont' get any ty error without

stelvio/component.py
```diff
@@ -245,6 +277,9 @@ async def handle_bridge_event(self, data: dict) -> BridgeInvocationResult | None
 class BridgeableMixin(ABC):
     _dev_endpoint_id: str | None = None
 
+    async def _handle_bridge_event(self, data: dict) -> BridgeInvocationResult | None:
+        raise NotImplementedError
+
```
michal-stlv: What's this for? we already have and and it's properly marked as `@abstract`

stelvio/aws/function/function.py
```diff
@@ -281,8 +277,8 @@ def _create_resources(self) -> FunctionResources:
             function_resource = lambda_.Function(
                 safe_name(context().prefix(), self.name, 64),
                 role=lambda_role.arn,
-                architectures=[function_architecture],
-                runtime=function_runtime,
+                architectures=[self.config.architecture] if self.config.architecture else None,
+                runtime=self.config.runtime,
```
michal-stlv: why architectures has if else while runtime does not?
also we do not run this thru customizer so if user does not provide value then runtime is None, but runtime is required value for so this will fail?

on unrelated note, theoretically for dev mode we can ignore user values and just go with arm and 128 as cheapest option since it's jus forwarding stuff but separate discusson, separate PR

stelvio/aws/dynamo_db.py
```diff
@@ -294,11 +294,15 @@ def _create_resources(self) -> DynamoSubscriptionResources:
...
-                    "starting_position": "LATEST",
-                    "batch_size": self._batch_size or 100,
+                    "batch_size": self._batch_size,
                     "maximum_batching_window_in_seconds": 0,
                     "filter_criteria": {"filters": self._filters} if self._filters else None,
                 },
+                default_props={
+                    "starting_position": "LATEST",
+                    "batch_size": 100,
+                    "maximum_batching_window_in_seconds": 0,
```
michal-stlv, line 303: this is also set in computed props, with same value (no user data) so should be only here not there?

michal-stlv (changes requested):
Left some comments for core logic and code stuff. 

Re tests:
- we dont' have test to test that explicit computed value beas a gloval callable's results? (although that behavour is questionable for some properties). the one thest we have passes `computed_props={"memory"...`
- `inject_tags`=True is never tested together with default_props and computed_props


I'd also get rid of typing stuff as it's for longer discussion so should be separate PR after we agree on the rules and ty settings.

stelvio/aws/function/function.py
sebst: good point, I'm just hardcoding the values for dev mode. would be justified to do this in this pr to avoid customizer call for lambda.

stelvio/component.py
sebst: > Is that correct?

yes

> user can't extent some props globally - e.g. adding some envar to all functions

adding env vars globally would not work (as we shallow merge later in the process)

> user actually don't have access to all props that component/resource is gonna use 

would add computed props as optional second argument to callable. not sure if that's not more confusing...

stelvio/aws/dynamo_db.py
sebst: good catch. fixed

stelvio/component.py
sebst: missed while removing intermediary code. removed.

stelvio/component.py
sebst: reverted typing

stelvio/component.py
sebst: reverted typing

stelvio/aws/function/function.py
sebst: renamed to props

docs/changelog.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## 0.10.0b6 (2026-MM-DD)
+
+### Component Customization
+
+- Customization values can now be **callables** that receive the resource's props and return the props to use, enabling dynamic customization.
+- **Global `customize` now acts as defaults, not overrides.** A global customization fills in Stelvio's defaults but no longer overrides values you set explicitly on a component. For example, with a global `memory=512` for all functions, `Function("my-fn", handler="...", memory=1024)` now deploys with `1024`.
```
michal-stlv: I don't think it's clear. Here. we should either explain briefly that dicts are defaults and callables are overrids. or just say we have callables and behavour is explained in the docs.

stelvio/aws/function/function.py
```diff
@@ -281,12 +278,12 @@ def _create_resources(self) -> FunctionResources:
...
-                runtime=function_runtime,
+                architectures=[DEFAULT_ARCHITECTURE_DEVMODE],
+                runtime=DEFAULT_RUNTIME,
                 code=_create_lambda_bridge_archive(),
                 handler="stlv_function_stub.handler",
                 environment={"variables": env_vars},
-                memory_size=self.config.memory or DEFAULT_MEMORY,
+                memory_size=DEFAULT_MEMORY,
                 timeout=self.config.timeout or DEFAULT_TIMEOUT,
                 layers=[layer.arn for layer in self.config.layers] if self.config.layers else None,
```
michal-stlv, line 288: I think layers are not needed since code is executing on local computer.  so we can strip them in dev mode. actually we might need to do it b/c f mismatch between layer architecture and lambda architecture.

however I'm not sure we document properly that user would need to make sure that locally their imports will work .

stelvio/component.py
```diff
@@ -109,7 +110,7 @@ def _get_valid_customize_keys(self) -> set[str] | None:
         Uses __annotations__ directly to avoid forward reference resolution issues.
         """
         # Walk up the MRO looking for Component with type args
-        for base in get_original_bases(type(self)):
+        for base in type(self).__orig_bases__:
```
michal-stlv: why this? on main branch we have `get_original_bases` which is idiomatic way to do it.

stelvio/component.py
```diff
@@ -328,7 +358,7 @@ def get_component_by_name(cls, name: str) -> Component[Any, Any] | None:
 
 
 def link_config_creator[T: Component](
-    component_type: type[T],
+    component_type: type[Component],
```
michal-stlv: All these type changes are wrong and differ from main.

stelvio/component.py
```diff
@@ -170,62 +171,91 @@ def _resource_opts(
...
+        *,
+        inject_tags: bool = False,
     ) -> dict[str, Any]:
-        """Apply global and per-instance customizations to default props.
+        """Apply global and per-instance customizations to resource props.
+
+        Global customize acts as *defaults*: it fills in or overrides Stelvio's
+        built-in defaults but does not silently override values set explicitly on
+        the component. Per-instance customize is applied last and overrides
+        everything.
```
michal-stlv: This is not correct. For global callables it is override - whatever it returns is used. Unless override with local customize. Behaviour is different thaan global dicts.

stelvio/component.py
```diff
@@ -170,62 +171,91 @@ def _resource_opts(
...
-            1. Per-instance customize (self._customize)
-            2. Global customize from StelvioAppConfig
-            3. Stelvio defaults (default_props)
-
-        Ordering details:
-            - Global customization is applied first.
-            - Local callable customization receives props after global
-              customization is already applied.
+            1. Per-instance customize (`self._customize`)
+            2. Explicit (non-`None`) values in `computed_props`
```
michal-stlv: only if not set by global callable - there callable decides if it will respect explicit value passed to instance

stelvio/component.py
```diff
@@ -27,7 +26,7 @@
 class Component[ResourcesT, CustomizationT](pulumi.ComponentResource, ABC):
     _name: str
     _resources: ResourcesT | None
-    _customize: CustomizationT
+    _customize: CustomizationT | None = None
```
michal-stlv: we dont' set default value here. this is just type anotation. we set value in `__init__` so no need for default also this defautl actually sets value for class variable not instance variable. so just be just `CustomizationT`. Actualy type before was correct because after init it's never None it's always dict.

stelvio/component.py
```diff
@@ -59,9 +58,11 @@ def __init__(
         super().__init__(type_name, name, None, resource_opts)
         self._name = name
         self._resources = None
-        self._customize = customize or {}
+        self._customize = customize
         self._tags = tags or {}
         self._validate_tags()
+        if self._customize is None:
+            self._customize = {}
```
michal-stlv: why not just one line `self._customize = customize or {}` instead of three?

tests/test_component.py
```diff
@@ -505,12 +506,59 @@ def local_customize(default_props):
...
+        "function",
+        computed_props={"memory": 1024, "timeout": None},
+        default_props={"memory": 128, "timeout": 10},
+    )
+
+    # Global customize acts as a default, while explicit computed props win.
+    assert result == {"memory": 1024, "timeout": 10}
+
+
+def test_customizer_explicit_computed_value_overrides_global_callable_default(clear_registry):
```
michal-stlv: test name here is misleading I'd say because wheter computed value overrides global callable is fully in the hands of global callable.

tests/test_component.py
```diff
@@ -557,6 +605,24 @@ def local_customize(props):
     assert "tags" not in result
 
 
+def test_customizer_inject_tags_with_computed_and_default_props(clear_registry):
+    component = MockComponent("tagged-resource", tags={"Team": "platform"})
+
+    result = component._customizer(
+        "resource",
+        computed_props={"name": "test", "memory": None, "tags": {"Env": "dev"}},
+        default_props={"memory": 128, "timeout": 30, "tags": {"Base": "yes"}},
```
michal-stlv, line 614: strange tests because tags would never be in default props really. but e.g. in vpc I do set lot of tags but those would end up in computed_props.  and those are merged by _customizer.

stelvio/aws/queue.py
```diff
@@ -72,19 +72,19 @@ class QueueConfig:
...
     visibility_timeout: int | None = None
-    retention: int = DEFAULT_QUEUE_RETENTION
+    retention: int | None = None
     dlq: Queue | str | DlqConfig | DlqConfigDict | None = None
 
 
 class QueueConfigDict(TypedDict, total=False):
     """Queue configuration dictionary."""
 
     fifo: bool
```
michal-stlv, line 84: if all others have `| None` why not this fifo?

tests/test_component.py
```diff
@@ -706,3 +772,417 @@ def test_resource_opts_defaults(clear_registry):
...
+    current_ctx = context()
+    _ContextStore.clear()
+    _ContextStore.set(
+        replace(
+            current_ctx,
+            customize={
+                MockComponent: {
+                    "function": lambda props: {
+                        **props,
+                        # Double the memory if it's None, otherwise double the provided default
```
michal-stlv: > Double the memory if it's None 

??? I think I know what you meant but if it's None and you double it it's still None. Nitpick but bit confusing , it's more like: if It's None double the default value

tests/test_component.py
```diff
@@ -706,3 +772,417 @@ def test_resource_opts_defaults(clear_registry):
...
+        "function",
+        computed_props={"memory": None},
+        default_props={"memory": 128, "timeout": 30},
+    )
+
+    # The new key is added on top of the defaults; memory falls back to its default.
+    assert result == {"memory": 128, "timeout": 30, "reserved_concurrent_executions": 5}
+
+
+def test_customizer_global_callable_changes_default_dynamically(clear_registry):
```
michal-stlv, line 1022: interesting tests but it's not really testing stelvio logic rather just global customize function?

tests/test_component.py
```diff
@@ -706,3 +772,417 @@ def test_resource_opts_defaults(clear_registry):
...
+        "function",
+        computed_props={"memory": 1024, "timeout": None},
+        default_props={"memory": 128, "timeout": 30},
+    )
+
+    # Local callable output: 4096 is highest precedence
+    assert result == {"memory": 4096, "timeout": 30}
+
+
+def test_customizer_multiple_resources_independent_customization(clear_registry):
```
michal-stlv: What are we really testing here? we only call instance variable so unless test Python implementation works ok with inherited methods  there's not much value to this?

michal-stlv (changes requested):
- Logic is correct and follows what we discussed. 

- I left few more code level comments. 

- Regarding tests I think all of them follow few patterns. we have some global customizer, some local. some defautls some computed and some expected result. so essentially all of that setup code etc. could be folded ino one or few helper functions that would take some params .  and then we could just have one parametrized test?

- We moved some properties to `default_props` variables but not all of those are thing sexposed as params to component init or fields of components config. I'm wondering if this need to be documented somehow? Maybe each component should list default props that are just static values (rather than computed).

stelvio/component.py
sebst: fixed

docs/changelog.md
sebst: hope it's clearer now?

stelvio/aws/function/function.py
sebst: that'd be a follow up ticket: #216

stelvio/component.py
sebst: fixed

stelvio/component.py
sebst: should be clearer now

stelvio/component.py
sebst: should be clearer now

stelvio/component.py
sebst: fixed

stelvio/component.py
sebst: fixed

tests/test_component.py
sebst: renamed

tests/test_component.py
sebst: clearer

tests/test_component.py
sebst: yes, just global customize function

tests/test_component.py
sebst: removed

stelvio/aws/queue.py
sebst: As a toggle (bool) it does make sense to have it as a default in `QueueConfig` (`fifo: bool = False`).

customizer would still work, and there's a limited use case to suddenly switch all Queues to fifo per global customizer, so this way we surface the default earlier to the user

tests/test_component.py
sebst: leave it as is: It tests the merge in the mock. I still think this test is valid. These mock component tetsts do not need to reflect the exact same things as the real components, but test primarily internal behavior.

stelvio/aws/queue.py
michal-stlv: i agree with logic but this comment is not about logic behind bool bur rather about if None makes sense in TypedDict with total=False. and we discussed that probably it's not needed. (as total false means value can be ommited) therefore its not really needed for others (in typed dict) or if it's there then for consistency should be also on fifo. but really it's a nit pick so you can ignore.

sebst:
> * Regarding tests I think all of them follow few patterns. we have some global customizer, some local. some defautls some computed and some expected result. so essentially all of that setup code etc. could be folded ino one or few helper functions that would take some params .  and then we could just have one parametrized test?

Prepared that. Would do that in separate PR.

> * We moved some properties to `default_props` variables but not all of those are thing sexposed as params to component init or fields of components config. I'm wondering if this need to be documented somehow? Maybe each component should list default props that are just static values (rather than computed).

Follow up ticket: #217

docs/changelog.md
```diff
@@ -4,8 +4,8 @@
 
 ### Component Customization
 
-- Customization values can now be **callables** that receive the resource's props and return the props to use, enabling dynamic customization.
-- **Global `customize` now acts as defaults, not overrides.** A global customization fills in Stelvio's defaults but no longer overrides values you set explicitly on a component. For example, with a global `memory=512` for all functions, `Function("my-fn", handler="...", memory=1024)` now deploys with `1024`.
+- Customization values can now be **callables** that receive the resource's props and return modified props, enabling dynamic customization based on component properties.
+- **Global `customize` now acts as defaults, not overrides.** Component-level values take precedence over global defaults. For example, with a global `memory=512` for all functions, `Function("my-fn", handler="...", memory=1024)` deploys with `1024`.
```
michal-stlv: It's still confusing. it says global customize act as defaults and component level takes precedence which is jut not true for callables.

stelvio/component.py
```diff
@@ -27,7 +27,7 @@
 class Component[ResourcesT, CustomizationT](pulumi.ComponentResource, ABC):
     _name: str
     _resources: ResourcesT | None
-    _customize: CustomizationT
+    _customize: CustomizationT | None
```
michal-stlv: it will never be None after init

tests/integration/run_all.sh
```diff
@@ -22,16 +22,16 @@ pids=()
 exit_code=0
 
 # Standard tier — 10 workers
-uv run pytest "$INTEGRATION_DIR" --integration $COMMON_ARGS -n 10 &
+uv run pytest "$INTEGRATION_DIR" --integration $COMMON_ARGS -n 10 -p no:warnings &
```
michal-stlv: what are this no:warnings?

michal-stlv (commented):
two small things and one question. otherwise good, given that tests rewrite and docs for defautls will be separate PR.

docs/changelog.md
michal-stlv: not really. when saying default we should be clear it's dicts only and callables behave differently.

stelvio/aws/function/function.py
michal-stlv: ok as far as dev mode works if layer is in x86 architecture and lambda in arm.

tests/test_component.py
michal-stlv: but then it doesn't need to exist?

stelvio/component.py
sebst: removed `| None`

docs/changelog.md
sebst: explicitly mentioned it now

tests/integration/run_all.sh
sebst: should've been a separate commit, so removed it here. this prevents lots of pulumi sdk warnings to surface during integration test runs.

michal-stlv (approved):

PR #213, opened 2026-06-21
fix: link type annotations

michal-stlv:


stelvio/link.py
```diff
@@ -1,54 +1,55 @@
...
 
-# Link has permissions, and each permission has actions and resources
+# Link has permissions, and each permission has actions and resources,
 # so permission represents part of statement
 @final
 @dataclass(frozen=True)
 class LinkConfig:
     properties: dict[str, Input[str]] | None = None
-    permissions: Sequence[AwsPermission] | None = None
+    permissions: list[AwsPermission] | None = None
```
sebst, line 33: why `Sequence` -> `list`?

stelvio/link.py
```diff
@@ -1,54 +1,55 @@
...
 
 
 @final
 @dataclass(frozen=True)
 class Link:
     name: str
     properties: dict[str, Input[str]] | None
-    permissions: Sequence[AwsPermission] | None
-    component: Optional["Component"] = None
+    permissions: list[Permission] | None
```
sebst, line 41: why `Sequence` -> `list`?

michal-stlv: because it is actually a list. sequence is more general so theoretically it could be something else. it ended up here  b/c pulumi api returns Sequence. but then we got some warnings in e.g. `return self.with_permissions(*(current + list(extra_permissions)))`. 
we can still keep `Sequence` if we really want for some good reason. but I just miss a good specific reason other then "let's keep it more general". on stelvio's side it's actually a list. 
applies to your other comment too.

sebst: thx

sebst (approved):

PR #214, opened 2026-06-22
docs: update customization guide (Callable support, default logic)

sebst:
…ult behavior


docs/docs/concepts/customization.md
```diff
@@ -15,7 +15,7 @@ Use the `customize` parameter when you need to:
 
 ## Basic Usage
 
-Pass a `customize` dictionary to any Stelvio component. The dictionary keys correspond to the underlying resources that the component creates:
+Pass a `customize` dictionary or callable to any Stelvio component. The dictionary keys correspond to the underlying resources that the component creates:
```
michal-stlv: you always pass dictionary. difference if value per key is dict or callable

docs/docs/concepts/customization.md
```diff
@@ -32,6 +32,23 @@ def run() -> None:
 
 In this example, `"bucket"` refers to the S3 bucket resource created by the `Bucket` component, and `force_destroy` is a Pulumi property that allows the bucket to be deleted even when it contains objects.
 
+You can also pass a callable to compute properties dynamically. It receives the resource's properties and returns the properties to use — see [Using Callables for Dynamic Customization](#using-callables-for-dynamic-customization) for how this differs per-instance vs. globally:
```
michal-stlv: > It receives the resource's properties

only computed. so not all. maybe properties different from defaults or something. as in prev PR comment maybe worth documenting default props

docs/docs/concepts/customization.md
```diff
@@ -120,19 +137,27 @@ table.subscribe(
...
 ## How Customization Works
 
-When you provide customizations, Stelvio merges your values with its default configuration:
+When you provide customizations, Stelvio applies them in this order (highest to lowest precedence):
 
-1. **Stelvio defaults** are applied first
-2. **Your customizations** override or extend those defaults
+1. **Per-instance customize** - Customizations passed directly to a component instance
+2. **Explicit values** - Properties explicitly set on the component (not None)
+3. **Global customize** - Customizations from `StelvioAppConfig` (acts as defaults)
```
michal-stlv, line 149: if it's callable it's override (unless overriden by local customizer), what it returns that's used (except none values)

docs/docs/concepts/customization.md
```diff
@@ -120,19 +137,27 @@ table.subscribe(
...
+4. **Stelvio defaults** - Built-in Stelvio default values
 
-This means you only need to specify the properties you want to change—Stelvio's sensible defaults remain in place for everything else.
+This means:
+- Explicit values you set always take precedence over global defaults
+- Global customize only applies if you don't set an explicit value
```
michal-stlv: not true for callable global customizer

docs/docs/concepts/customization.md
```diff
@@ -120,19 +137,27 @@ table.subscribe(
...
+- Per-instance customize overrides everything
+- Stelvio's sensible defaults remain in place for properties you don't customize
+
+Global *callables* are the exception: they receive the props and can transform or even override them. See [Using Callables for Dynamic Customization](#using-callables-for-dynamic-customization).
```
michal-stlv: ok I see this fixes my previous comments, still maybe it should be sooner or explain upfornt that callables behave differently. People might not read till the end :)

docs/docs/concepts/customization.md
```diff
@@ -120,19 +137,27 @@ table.subscribe(
...
 
 !!! note "Shallow Merge"
-    The merge is shallow at each property level. If you customize a nested object (like `tags`), 
+    The merge is shallow at each property level. If you customize a nested object,
```
michal-stlv, line 163: we should state merge applies only to dicts and not callablesl

docs/docs/concepts/customization.md
```diff
@@ -141,57 +166,62 @@ This means you only need to specify the properties you want to change—Stelvio'
...
         }
     }
 )
 ```
 
-#### Global + Instance Tag Replacement
+#### Explicit Values Override Global Defaults
 
-The same shallow merge applies when combining global and per-instance customization:
+With the new behavior, explicit values take precedence over global defaults:
```
michal-stlv: same thing as before. only dicts.

docs/docs/concepts/customization.md
```diff
@@ -141,57 +166,62 @@ This means you only need to specify the properties you want to change—Stelvio'
...
+    # ✅ Explicit value overrides global default: memory_size = 1024
+    fn2 = Function(
+        "fn2",
+        handler="handlers.handler",
+        memory=1024,  # Explicit value takes precedence
     )
-    # Result: tags = {"Env": "dev"} - Team and Cost are gone!
 ```
 
+This is the key difference from the old behavior: you no longer need to use `customize` to override global defaults—explicit constructor arguments work naturally.
```
michal-stlv: do we need to mention old behavior? just state how it is now?

docs/docs/concepts/customization.md
```diff
@@ -141,57 +166,62 @@ This means you only need to specify the properties you want to change—Stelvio'
...
+
+!!! note "`memory` vs. `memory_size`"
```
michal-stlv: good point, maybe also worth general note that stelvio properties names do not map 1:1 to pulumi properties. (and details in default props docs I mentioned)

docs/docs/concepts/customization.md
```diff
@@ -141,57 +166,62 @@ This means you only need to specify the properties you want to change—Stelvio'
...
+    Stelvio's `memory` constructor argument maps to the underlying Pulumi
+    `memory_size` property — the name you use inside `customize`.
+
 ## Global Customization
 
-Apply customizations to all instances of a component type using the `customize` option in `StelvioAppConfig`:
+Apply default customizations to all instances of a component type using the `customize` option in `StelvioAppConfig`. Global customizations act as **defaults**—explicit values in component constructors override them:
```
michal-stlv, line 234: again only for dicts

michal-stlv (changes requested):
Left some comments mostly around defaults vs. overrides in the context dict/callable. 

Maybe we could restrucure docs somehow. 

e.g. Stelvio suports customization with simple dicts and callable. 

and then have two sections one for dicts and one for callable....

not saying it's best solution and I'm open to other suggestions but now it's confusing as you mention that global is default but then readre can be confused as you dont' say in all those places that it's for dicts only.

michal-stlv (approved):

PR #218, opened 2026-07-03
chore: parameterize customizer tests

sebst:


tests/test_component.py
```diff
@@ -295,71 +295,173 @@ def test_creator(r):
...
+    ],
+)
+def test_customizer_dict_patterns(
+    clear_registry, resource_name, local_customize, default_props, expected
+):
+    """Parametrized test for dict-based customization patterns."""
+    if local_customize is not None:
+        component = MockComponent("test-component", customize=local_customize)
+    else:
+        component = MockComponent("test-component")
```
michal-stlv: why not just 
```py
component = MockComponent("test-component", customize=local_customize)
```

None is valid value, default even.

michal-stlv (commented):
left one comment/question.

also most of parametrized test have no ids. They have helpful comments which is nice:
```py
# Resource not in customize
        (
            "some_resource",
            {"other_resource": {"key1": "override1"}},
            {"key1": "value1", "key2": "value2"},
            {"key1": "value1", "key2": "value2"},
        ),
```

but that doesn't show in runner.

maybe we could do:
```py
from pytest import param

...
        param(
            "some_resource",
            {"other_resource": {"key1": "override1"}},
            {"key1": "value1", "key2": "value2"},
            {"key1": "value1", "key2": "value2"},
            id="resource-not-in-customize"
        ),
```

michal-stlv (approved):

PR #222, opened 2026-07-16
feat: add `HttpApi` component

sebst:


stelvio/aws/api_gateway/http_api/domain.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class ApiDomainCustomizationDict(TypedDict, total=False):
+    certificate: pulumi_aws.acm.CertificateArgs | dict[str, Any] | None
+    domain: pulumi_aws.apigatewayv2.DomainNameArgs | dict[str, Any] | None
+    dns_record: dict[str, Any] | None
+
+
+@final
+class ApiDomain(Component[HttpApiDomainResources, ApiDomainCustomizationDict]):
```
michal-stlv: we have `ApiDomain` but `HttpApiDomainResources`  ?

stelvio/aws/api_gateway/http_api/http_api.py
```diff
@@ -0,0 +1,808 @@
...
+    routes: list[apigatewayv2.Route]
+    permissions: list[lambda_.Permission]
+    api_mapping: apigatewayv2.ApiMapping | None = None
+
+
+class HttpApiCustomizationDict(TypedDict, total=False):
+    api: apigatewayv2.ApiArgs | dict[str, Any] | None
+    stage: apigatewayv2.StageArgs | dict[str, Any] | None
+    log_group: cloudwatch.LogGroupArgs | dict[str, Any] | None
+    api_mapping: apigatewayv2.ApiMappingArgs | dict[str, Any] | None
```
michal-stlv, line 124: I still don't get why `apigatewayv2.Api` and `cloudwatch.LogGroupArgs` is better than `api: apigatewayv2.Api` and `from pulumi.cloudwatch import LogGroupArgs` . I will not block PR on this but kills my eyes and take longer to read. applies to other places and files in this PR. i mean makes sense sometimes if name conflict otherwise it's noise that takes space.

michal-stlv (changes requested):

stelvio/aws/api_gateway/http_api/domain.py
sebst: renamed

stelvio/aws/api_gateway/http_api/http_api.py
sebst: would add these kind of improvements to stack  5

michal-stlv (approved):

stelvio/aws/cors.py
```diff
@@ -36,6 +38,35 @@ class CorsConfigDict(TypedDict, total=False):
     expose_headers: list[str] | None
 
 
+def normalize_cors_config(
```
michal-stlv, line 41: any reason why this is moved up before CorsConfig definition?

michal-stlv (approved):

PR #224, opened 2026-07-18
feat: add tests for `HttpApi`

sebst:


tests/aws/api_gateway/http_api/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+from stelvio.aws.api_gateway.http_api import HttpApi
+
+from ...pulumi_mocks import PulumiTestMocks
+
+TP = "test-test-"
+
+
+def reset_caches() -> None:
+    """Clear cached IAM role creation for API Gateway."""
+    from stelvio.aws.api_gateway.rest_api.iam import _create_api_gateway_account_and_role
```
michal-stlv, line 12: I've just noticed we use `_create_api_gateway_account_and_role` in http and rest apis but it's in rest_api module. Should be moved to a common one under api_gateway. It's more of PR 2 territory but should be done here or there.

tests/aws/api_gateway/http_api/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+    outputs = [resources.stage.id]
+    outputs.extend(permission.id for permission in resources.permissions)
+    outputs.extend(route.id for route in resources.routes)
+    if resources.api_mapping is not None:
+        outputs.append(resources.api_mapping.id)
+    pulumi.Output.all(*outputs).apply(callback)
+
+
+@pytest.fixture
+def pulumi_mocks() -> PulumiTestMocks:
```
michal-stlv: this as well as `reset_caches` are same as in `rest_api/conftest.py`. I'd say we can have a conftest right under api_gateway shared by both rest/http and maybe also websockets later, removing duplications of `pulumi_mocks` and `reset_caches`

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,795 @@
...
+
+
+def test_http_api_invalid_stage_name_raises():
+    with pytest.raises(ValueError, match="Stage name"):
+        HttpApi("my-api", stage_name="$bad")
+
+
+def test_http_api_invalid_log_retention_raises():
+    with pytest.raises(ValueError, match="access_log_retention_days"):
+        HttpApi("my-api", access_log_retention_days=999)
```
michal-stlv: These could be one parametrized test

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,795 @@
...
+@pytest.mark.parametrize("days", [1, 30, 3653])
+def test_http_api_access_log_retention_valid_boundaries(days):
+    # Must not raise
+    HttpApiConfig(access_log_retention_days=days)
+
+
+@pytest.mark.parametrize("days", [0, 2, 9999])
+def test_http_api_access_log_retention_invalid_boundaries(days):
+    with pytest.raises(ValueError, match="access_log_retention_days"):
+        HttpApiConfig(access_log_retention_days=days)
```
michal-stlv: these could be tested directly with `HttpApi` rather than config: `HttpApi("my-api", access_log_retention_days=days)` also we don't check that any custom value also ends up in the actual log group (only the default 30).

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,795 @@
...
+
+def test_http_api_link_config_structure():
+    from stelvio.component import ComponentRegistry
+
+    api = HttpApi("orders")
+    api.route("GET", "/orders", "functions/simple.handler")
+    creator = ComponentRegistry.get_link_config_creator(HttpApi)
+    link = creator(api)
+    assert set(link.properties) == {"api_url", "api_execution_arn"}
+    assert link.permissions == []
```
michal-stlv: This goes through `ComponentRegistry.get_link_config_creator` to inspect the link config directly. The properties half is already covered by `test_http_api_link_injects_api_url_env_vars`. The `permissions == []` half is the only new information, but the actual thign we test is "linking an HttpApi doesn't add any IAM policies to the function's role"? could we assert that in the env-var test instead (e.g. no role policies created for the client function) and delete this one?

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,795 @@
...
+        "my-api",
+        domain=domain,
+        api_mapping_key="v2",
+        disable_execute_api_endpoint=True,
+    )
+
+    assert api.config.domain_name is None
+    assert api.domain_name == "api.example.com"
+    assert api.config.api_mapping_key == "v2"
+    assert api.config.disable_execute_api_endpoint is True
```
michal-stlv: I'm not sure what we test here. To me only meaningful thing here is `assert api.domain_name == "api.example.com"` so we prove that `domain_name` is set from shared ApiDomain. rest of the asserts just test we pass params to config?

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,795 @@
...
+    api.route(["GET", "DELETE"], "/users/{id}", "functions/simple.handler")
+    _ = api.resources
+
+    def check(_):
+        routes = pulumi_mocks.created_http_api_routes()
+        route_keys = {r.inputs["routeKey"] for r in routes}
+        assert route_keys == {"GET /users/{id}", "DELETE /users/{id}"}
+
+    when_http_api_ready(api, check)
+
```
michal-stlv, line 460: These could also be one parametrized test.

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,795 @@
...
+@pulumi.runtime.test
+def test_http_api_url_named_stage_execute_api(pulumi_mocks):
+    api = HttpApi("my-api", stage_name="prod")
+    api.route("GET", "/users", "functions/simple.handler")
+
+    def check(urls):
+        assert urls[0] == urls[1]
+
+    pulumi.Output.all(api.url, api.resources.stage.invoke_url).apply(check)
+
```
michal-stlv, line 644: when there's no custom domain, `api.url` just returns `stage.invoke_url`, so these tests compare the same value with itself, they can't really fail? I mean only if we break that url returns invoke_url. But can we assert the actual expected URL instead, e.g. `https://{HTTP_API_ID}.execute-api.us-east-1.amazonaws.com/prod` for the named-stage test? The mocks build invoke_url from the stage's real inputs, so this would also catch the stage name not making it into the URL, right now nothing checks the /prod suffix.

tests/aws/api_gateway/http_api/test_http_api_route_dataclass.py
```diff
@@ -0,0 +1,171 @@
+"""Direct unit tests for the `_HttpRoute` dataclass.
+
+`_HttpRoute` is the validation/normalization layer between `HttpApi.route()` and
+resource creation. Testing it directly keeps these concerns cheap and focused,
+independent of the Pulumi mock runtime.
```
michal-stlv: Yes and I know we do the same in RestApi but I'd actually prefer to delete this file or rename and write the same tests in more behavior way one level up using `HttpApi` directly.

Why: `_HttpRoute` is private, so if we ever rename it or merge it into `routes.py` all 9 tests break with zero behavior change. And going through `api.route()` costs nothing: validation raises at call time, so the error tests need no Pulumi runtime or mocks, they're exactly as fast as they are now.

How I'd do it:

- **methods normalization + route keys**: these are really one behavior, which routes end up created. Make one parametrized test that calls `api.route(method, path, handler)` and asserts the created routes' `routeKey` inputs. The 4 single tests in `test_http_api.py` (`ANY`, `*`, `$default`, multi-method) can become rows in the same table. Make sure to keep the lowercase `"get"` and enum rows from here, those aren't covered anywhere else right now.
- **validation errors**: `test_http_api_routes.py` already has the same table via `api.route()`. Just add the rows that only exist in this file: path too long, more than 10 path params, empty method list, plus the max-boundary accept case. The invalid method type cases stay their own small table since they're `TypeError`, not `ValueError`.
- **invalid handler types**: double-check what `api.route("GET", "/x", 123)` actually raises. It goes through `parse_handler_config` before `_HttpRoute`, so it might not be this `TypeError` at all. Whatever the user actually sees is what we should pin.
- **`Function` instance / `FunctionConfig` preserved**: already covered in `test_http_api_route.py` (supplied `Function`'s name shows up on the lambda, `memory=512` shows up in inputs). Just delete these two.
- **`jwt_scopes` stored-not-validated**: delete. It asserts internal state; the real behavior (rejection at resource creation) is already covered by the `test_jwt_scopes_with_*` tests.

I'll put the RestApi equivalents (`test_api_route_dataclass.py`, `test_api_helper_functions.py`) on the cleanup list separately, no need to touch them in this PR.

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,795 @@
+"""Unit tests for HttpApi component (API Gateway v2)."""
```
michal-stlv, line 1: Bigger structural thing, applies to this whole file and also the other new http_api test files (and some of my other comments here, the "could be one parametrized test" ones would be covered by this too): in the VPC tests (`tests/aws/test_vpc.py`) I added new mock assert helpers, `assert_res`, `assert_res_counts` and the `R` enum. I'd like this suite to use them instead of `created_*` + counting by hand.

Right now every test checks a small slice, counts live only in the first test, the role/account asserts repeat, and nothing is sealed, so if we create some extra resource by accident no test fails. What I'd like:

- rebase the stack onto main first, the helpers are not on this branch yet
- a `verify_http_api(mocks, tc)` similar to RestApi's `assert_api_gateway_resources` (which I think gets it right, every test verifies the whole graph, not a slice): assert the apigatewayv2 resources, permissions and shared role/account fully with `assert_res`, seal with `assert_res_counts`
- functions only partially (`partial=True`): `handler` plus what the case sets (`memorySize`, `timeout`). The `Function` suite owns the rest. Don't do it like RestApi's `assert_stelvio_functions` which reads `ComponentRegistry._instances`, assert the created lambda resources' inputs instead, that checks what actually got deployed
- drive it with test cases like `VpcTestCase`: inputs (routes, `stage_name`, `cors`, retention, domain) and literal expected values (route keys, integration URIs, permission source ARNs). One parametrized `test_http_api` can then replace the first test, the per-resource tests, multi-route, shared handler, stage name, cors and retention tests
- keep the small single tests that are not "deploy and check the graph": validation raises, freeze after create, `url` property, link env vars

I let Claude prototype this to check it makes sense: 7 cases replaced 14 tests (268 vs 310 lines) and every case now checks the whole graph with sealed counts, so variants like custom stage or cors get full verification, not just the one changed field. So better coverage with a bit less code.

I know it's not a small thing but I'd like this in this PR, not as a follow-up. The rebase is needed anyway, and otherwise we merge 2k lines of a style we're moving away from plus new `created_http_api_*` helpers in `pulumi_mocks.py` that would become dead code right after. Happy to jump on a call to discuss.

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,795 @@
...
+
```
michal-stlv, line 2: Separate thing, missing coverage. I don't see tests for:

- the implicit domain path: `HttpApi("x", domain_name="api.example.com")` creating its own `ApiDomain`. Nothing asserts the created `DomainName`, the root `ApiMapping` and the DNS record, all domain tests here go through a shared `ApiDomain` or only check the `api` inputs
- a mapping without `api_mapping_key`, root mapping, `apiMappingKey` absent from the inputs
- `config=123` raising `TypeError`

tests/aws/api_gateway/http_api/test_http_api_config.py
```diff
@@ -0,0 +1,51 @@
+import pytest
```
michal-stlv: Same thing as `test_http_api_route_dataclass.py` but smaller: `HttpApi(**opts)` goes straight to `HttpApiConfig`, so all of these can be tested through `HttpApi("my-api", stage_name="with spaces")` etc. Some rows are already covered there (`api_mapping_key` without domain, `disable_execute_api_endpoint` without domain, bad stage name), so after moving the unique ones (empty/whitespace domain, spaces and 129 char stage names, `domain_name` + `domain` conflict, the domain type errors) this file only needs to keep `test_http_api_config_dict_matches_http_api_config`.

`test_http_api_config_defaults` can go too, the defaults that matter are already checked on the created resources (stage `$default`, retention 30, no cors config).

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
```diff
@@ -0,0 +1,586 @@
...
+        assert authorizers[0].inputs["enableSimpleResponses"] is expected_simple_response
+        assert authorizers[0].inputs["identitySources"] == ["$request.header.Authorization"]
+        # Authorizer URI must be the invoke ARN of its dedicated Lambda
+        expected_uri = LAMBDA_INVOKE_ARN_TEMPLATE.format(
+            function_name=tn(TP + "my-api-auth-my-auth")
+        )
+        assert authorizers[0].inputs["authorizerUri"] == expected_uri
+
+        # A Lambda function for the authorizer is registered with the right handler
+        funcs = {f.name: f for f in ComponentRegistry._instances.get(Function, [])}
```
michal-stlv: The `ComponentRegistry._instances` part can go, the `created_functions` asserts below already check the same thing on the actual resource. Also `expected_simple_response` just mirrors `simple_response`, one param is enough. And while here: we never assert the default `authorizerResultTtlInSeconds` (300) on the created authorizer, only `ttl=0`, would be good to add.

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
```diff
@@ -0,0 +1,586 @@
...
+    api = HttpApi("my-api")
+    with pytest.raises(TypeError, match="identity_sources"):
+        api.add_lambda_authorizer(
+            "my-auth",
+            "functions/simple.handler",
+            identity_sources="method.request.header.Authorization",
+        )
+
+
+def test_lambda_authorizer_supports_function_config_dict():
```
michal-stlv: This checks the authorizer object's internals (`auth.function.config`). The real behavior is that the dict becomes the auth lambda's config, so better to deploy and assert `created_functions(...)[0].inputs["timeout"] == 10`.

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
```diff
@@ -0,0 +1,586 @@
...
+        api.add_jwt_authorizer("jwt", issuer="", audiences=["aud"])
+
+
+def test_jwt_authorizer_empty_audiences_raises():
+    api = HttpApi("my-api")
+    with pytest.raises(ValueError, match="audiences"):
+        api.add_jwt_authorizer("jwt", issuer="https://example.com", audiences=[])
+
+
+def test_jwt_authorizer_keeps_identity_source_as_given():
```
michal-stlv: Same for the cognito one below: this asserts the attribute on the returned object, not what ends up in the resource. If we drop `identity_source` on the way into `apigatewayv2.Authorizer` this stays green. Assert `identitySources` on the created authorizer instead, then both tests can be one parametrized test.

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
```diff
@@ -0,0 +1,586 @@
+"""Tests for HTTP API authorizers."""
```
michal-stlv, line 1: Two error paths I don't see covered: `jwt_scopes=[""]` with a JWT authorizer (the non-empty strings check) and a `UserPoolClient` audience combined with an ARN string `user_pool` (should be `TypeError`).

Also the `assert_res` / `assert_res_counts` comment from `test_http_api.py` applies here too. I don't expect the `HttpApiTestCase` shape to fit authorizers 1:1, they probably need their own small case shape or just keep the targeted tests, but either way use `assert_res` with full inputs instead of `created_*` + checking a few keys, and seal the counts.

tests/aws/api_gateway/http_api/test_http_api_cors.py
```diff
@@ -0,0 +1,99 @@
...
+            "allowOrigins": ["https://app.example.com"],
+            "allowMethods": ["GET"],
+            "allowHeaders": ["Authorization"],
+        }
+
+    api.resources.api.id.apply(check)
+
+
+@pulumi.runtime.test
+def test_multiple_apis_with_cors_create_uniquely_named_resources(pulumi_mocks):
```
michal-stlv: This mostly repeats `test_multiple_apis_with_same_routes_coexist_with_unique_resource_names` from `test_http_api.py`, same uniqueness assert, same wait outputs block. The cors part doesn't add much, cors is one input on the Api resource and `test_http_api_cors_true` already covers it. I'd delete this one, or if you want to keep a cors flavor, add `cors=True` to the existing coexistence test.

tests/aws/api_gateway/http_api/test_http_api_cors.py
```diff
@@ -0,0 +1,99 @@
+import pulumi
```
michal-stlv: And same note as in `test_http_api.py`: the two remaining tests here fit the new style too, cors variants can be spec cases (or at least `assert_res` with the full Api inputs instead of checking just `corsConfiguration`). Same for asserts in `tests/aws/api_gateway/http_api/test_http_api_domain.py`

tests/aws/cloudfront/test_cloudfront_adapter_api_gateway.py
```diff
@@ -161,71 +159,31 @@ def test_adapter_with_various_indices(idx):
...
-            "            return request;\n"
-            "        }"
-        )
-
-    pulumi.Output.all(
-        origin_config.origins["origin_id"],
-        origin_config.origins["domain_name"],
-        origin_config.origins["origin_path"],
-        origin_config.cloudfront_functions.arn,
-    ).apply(check_resources)
```
michal-stlv: This removes the only test that actually checked the RestApi origin config (domain, `/v1` origin path, allowed/cached methods, TTLs, the uri rewrite function code) and replaces it with comments plus `assert adapter.api == mock_api`, which can't fail for any real reason. The "we can't test without Pulumi mocks" comment isn't right, `test_router_creates_cloudfront_origin_for_http_api` in this same PR does exactly that through `Router`. Please bring the coverage back as a Router level test for `RestApi`, same shape as the http_api one, and delete this placeholder.

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def test_http_api_adapter_is_registered():
+    CloudfrontAdapterRegistry._ensure_adapters_loaded()
+
+    mock_api = Mock(spec=HttpApi)
+    adapter_class = CloudfrontAdapterRegistry.get_adapter_for_component(mock_api)
+
+    assert adapter_class == HttpApiCloudfrontAdapter
+    assert HttpApiCloudfrontAdapter.component_class == HttpApi
```
michal-stlv: These and `test_http_api_adapter_access_policy_is_not_needed`  test the adapter class through mocks and registry internals, but the Router tests below already prove all of it: `router.route("/api", api)` producing an execute-api origin means the adapter is registered and matches, and `oacs == 0` covers `get_access_policy` returning None. I'd delete them (yes the rest_api adapter file has the same ones that should be follow up later).

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
```diff
@@ -0,0 +1,229 @@
...
+        assert len(apis) == 1
+
+        distributions = pulumi_mocks.created_cloudfront_distributions()
+        assert len(distributions) == 1
+        distribution = distributions[0]
+
+        origins = distribution.inputs["origins"]
+        assert len(origins) == 1
+        origin = origins[0]
+        assert origin["domainName"].endswith(".execute-api.us-east-1.amazonaws.com")
```
michal-stlv: The domain can be asserted exactly, the mock makes it `{tid(TP + "edge-api")[:8]}.execute-api.us-east-1.amazonaws.com`. And let's also pin `cachedMethods`, `minTtl` and `maxTtl` here, the old rest_api adapter test checked those and this is now the only place that covers this behavior.

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
```diff
@@ -0,0 +1,229 @@
...
+    router.route("/api", api)
+    resources = router.resources
+
+    def check(_):
+        distribution = pulumi_mocks.created_cloudfront_distributions()[0]
+        origin = distribution.inputs["origins"][0]
+        assert origin["domainName"] == "api.example.com"
+        assert origin["originPath"] == "/v1"
+
+    resources.distribution.id.apply(check)
```
michal-stlv: These are one parametrized test: (api kwargs) -> (expected `domainName`, `originPath`). Six tests, same arrange and assert. The default stage case is also already asserted inside `test_router_creates_cloudfront_origin_for_http_api`.

tests/aws/api_gateway/rest_api/test_rest_api.py
```diff
@@ -387,8 +387,8 @@ def assert_stelvio_functions(
 
 
 @pulumi.runtime.test
-def test_rest_api_properties(pulumi_mocks):
-    """Test that RestApi.resources property correctly provides access to created resources."""
+def test_api_properties(pulumi_mocks):
+    """Test that Api.resources property correctly provides access to created resources."""
```
michal-stlv: This rename goes the wrong direction, everything else in this PR renames `Api` -> `RestApi`. I guess accident?

tests/aws/test_customization_sync.py
```diff
@@ -151,7 +157,21 @@
             RestApiCustomizationDict,
             None,
             None,
-            id="RestApi",
+            id="Api",
```
michal-stlv: Same backwards rename as in `test_rest_api.py`, this should stay `id="RestApi"`.

stelvio/__init__.py
michal-stlv, line 1: Accidental chmod? The file mode changed to executable with no content change. mistake?

tests/integration/test_http_api.py
```diff
@@ -0,0 +1,308 @@
...
+        export_function(fn)
+        export_http_api(api)
+
+    outputs = stelvio_env.deploy(infra)
+
+    assert_http_api_routes(
+        outputs["http_api_sharedhttpapi_id"],
+        expected_route_keys={"GET /one", "POST /two"},
+    )
+    assert "function_sharedhttp_arn" in outputs
```
michal-stlv, line 307: This only checks the routes exist and the export key is there, nothing verifies both routes actually go to the shared function. Either check via boto3 that both integrations point at the exported function arn, or just hit both routes and check 200.

tests/integration/test_http_api.py
```diff
@@ -0,0 +1,308 @@
...
+    outputs = stelvio_env.deploy(infra)
+    api_id = outputs["http_api_iamhttp_id"]
+    base_url = outputs["http_api_iamhttp_url"]
+
+    assert_http_api_route_auth(api_id, route_key="GET /admin", auth_type="AWS_IAM")
+
+    time.sleep(_HTTP_API_DEPLOY_WAIT)
+    status, _ = http_request(f"{base_url}/admin")
+    # Unsigned request must be rejected (403 Forbidden / Missing Authentication Token)
+    assert status in (401, 403)
```
michal-stlv: HTTP API rejects unsigned requests with 403, we can pin `== 403`.

tests/integration/assert_helpers.py
```diff
@@ -723,6 +730,87 @@ def assert_api_method_auth(
...
+        if mapping["ApiId"] == expected_api_id
+        and mapping.get("ApiMappingKey") == expected_mapping_key
+    ]
+    assert len(matching) == 1, (
+        f"Expected one mapping for API {expected_api_id!r} and key "
+        f"{expected_mapping_key!r}, got {mappings}"
+    )
+
+
+def assert_http_api_cors_headers(invoke_url: str, path: str = "/") -> None:
```
michal-stlv: This is the same code as `assert_api_cors_headers` above (after the lowercase fix they're identical). One helper is enough, the URL scheme check works for both.

And while here: these only check the headers exist, not their values. With `cors=True` we know it's `*`, so let's assert the actual value.

tests/integration/export_helpers.py
```diff
@@ -43,10 +43,26 @@ def export_api(api: RestApi) -> None:
...
+    r = api.resources
+    export_output(f"http_api_{api.name}_arn", api.arn)
+    export_output(f"http_api_{api.name}_execution_arn", api.execution_arn)
+    export_output(f"http_api_{api.name}_id", api.api_id)
+    export_output(f"http_api_{api.name}_url", api.url)
+    export_output(f"http_api_{api.name}_stage_id", r.stage.id)
+    export_output(f"http_api_{api.name}_stage_name", r.stage.name)
+
+
+def export_http_api_domain(domain: ApiDomain) -> None:
```
michal-stlv, line 61: `export_http_api_domain` isn't used by any test. Either drop it or use it in `test_dns_http_api.py`, asserting `target_domain_name` there would actually be a nice addition.

tests/integration/conftest.py
```diff
@@ -23,10 +23,10 @@
 #   integration     — standard tests, AWS profile only (142 tests / 10 workers)
 #     STLV_TEST_AWS_PROFILE=<profile> uv run pytest tests/integration/ --integration -v -n 10
 #
-#   integration_cf  — CloudFront/Router/S3StaticWebsite, slow teardown (13 tests / 7 workers)
+#   integration_cf  — CloudFront/Router/S3StaticWebsite, slow teardown (14 tests / 7 workers)
```
michal-stlv, line 26: The cf and dns tier counts got updated but the standard tier didn't, this says 142, run_all.sh says 172, and the ~15 new tests here aren't counted in either. Please recount and update both, and check the worker split still comes out even.

michal-stlv (changes requested):
Great coverage overall, the wiring tests (integration URIs, route targets, permission ARNs) are exactly right. Two big asks in the comments: move the suite to the new `assert_res` / spec-case style from the VPC tests (prototype confirms it works, details in the `test_http_api.py` comment), and bring back the RestApi CloudFront origin coverage that this PR removes (`test_cloudfront_adapter_api_gateway.py`). Rest is smaller: a few files that test private helpers directly, should be rewritten through HttpApi level tests, some parametrize opportunities, two accidental renames and a chmod.

tests/aws/api_gateway/http_api/test_http_api.py
sebst: done

tests/aws/api_gateway/http_api/test_http_api.py
sebst: done

tests/aws/api_gateway/http_api/test_http_api.py
sebst: done

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
sebst: done

tests/aws/api_gateway/rest_api/test_rest_api.py
sebst: fixed

tests/aws/test_customization_sync.py
sebst: fixed

stelvio/__init__.py
sebst: fixed

tests/integration/test_http_api.py
sebst: fixed

tests/integration/conftest.py
sebst: fixed

tests/integration/export_helpers.py
sebst: done test_dns_http_api.py

tests/integration/assert_helpers.py
sebst: fixed

tests/integration/test_http_api.py
sebst: done

tests/aws/api_gateway/http_api/conftest.py
sebst: done

tests/aws/api_gateway/http_api/conftest.py
sebst: done

tests/aws/api_gateway/http_api/test_http_api.py
sebst: done

tests/aws/api_gateway/http_api/test_http_api.py
sebst: removed

tests/aws/cloudfront/test_cloudfront_adapter_api_gateway.py
sebst: test_router_creates_cloudfront_origin_for_rest_api

tests/aws/api_gateway/http_api/test_http_api.py
sebst: included /prod

tests/aws/api_gateway/http_api/test_http_api_route_dataclass.py
sebst: deleted test_http_api_route_dataclass.py

tests/aws/api_gateway/http_api/test_http_api.py
sebst: done HttpApiTestCase + verify_http_api

tests/aws/api_gateway/http_api/test_http_api.py
sebst: Added

tests/aws/api_gateway/http_api/test_http_api_config.py
sebst: done. dropped test_http_api_config_defaults

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
sebst: done simple_response

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
sebst: done

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
sebst: done

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
sebst: added

tests/aws/api_gateway/http_api/test_http_api_cors.py
sebst: deleted and cors=True on existing multi api test

tests/aws/api_gateway/http_api/test_http_api_cors.py
sebst: done assert_api_domain_graph / assert_api_mapping

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
sebst: deleted

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
sebst: done

tests/integration/conftest.py
sebst: 175 standard / 10 workers, 14 CloudFront / 7, 9 DNS / 3

sebst:
> Great coverage overall, the wiring tests (integration URIs, route targets, permission ARNs) are exactly right. Two big asks in the comments: move the suite to the new `assert_res` / spec-case style from the VPC tests (prototype confirms it works, details in the `test_http_api.py` comment), and bring back the RestApi CloudFront origin coverage that this PR removes (`test_cloudfront_adapter_api_gateway.py`). Rest is smaller: a few files that test private helpers directly, should be rewritten through HttpApi level tests, some parametrize opportunities, two accidental renames and a chmod.

fixed

PR #225, opened 2026-07-18
refactor: remove redeploy() alias from StelvioTestEnv

michal-stlv:


sebst (approved):
why tho?

michal-stlv:
> why tho?

Because it did nothing ?

PR #238, opened 2026-07-24
feat: http api tests and docs

### LINKED ISSUE #215: Support reuse of domain among rest and http apis
It should be possible to configure `pulumi_aws.apigatewayv2.DomainName` so it's also usable with rest api and also websockets so users could share one domain among http and rest apis (as far as they're regional) and websockets. 

Then we could also have `RestApiDomain` (using `apigateway.DomainName`) that would handle the case where user needs to share domain between rest apis which are `edge`.

sebst:
closes #215


stelvio/aws/api_gateway/__init__.py
```diff
@@ -24,12 +24,20 @@
...
+    "HTTPMethod",
+    "HttpApi",
     "HttpApi",
     "HttpApiConfig",
+    "HttpApiConfig",
+    "HttpApiConfigDict",
     "HttpApiConfigDict",
     "HttpApiCustomizationDict",
+    "HttpApiCustomizationDict",
+    "HttpApiResources",
```
michal-stlv: 8 duplicate entries here

update: fixed in c28c66a already, nothing to do here.

tests/aws/api_gateway/http_api/conftest.py
```diff
@@ -0,0 +1,42 @@
...
+
+    if hasattr(_create_api_gateway_account_and_role, "cache_clear"):
+        _create_api_gateway_account_and_role.cache_clear()
+
+
+@pytest.fixture(autouse=True)
+def reset_api_gateway_caches():
+    reset_caches()
+    yield
+    reset_caches()
```
michal-stlv: Cache-clear belongs one level up. Here it only covers http_api both ways. rest_api tests clear just at setup, and cloudfront/tagging tests create APIs with no clear at all, so rest_api then cloudfront ordering can still flake. And http_api tests now clear 3x per test.

One autouse fixture in tests/aws/conftest.py, everything else goes:
```python
from stelvio.aws.api_gateway.iam import _create_api_gateway_account_and_role

@pytest.fixture(autouse=True)
def reset_api_gateway_cache():
    _create_api_gateway_account_and_role.cache_clear()
    yield
    _create_api_gateway_account_and_role.cache_clear()
```

Delete reset_caches + reset_api_gateway_caches here. The `hasattr` guard is dead code, and the import points at rest_api.iam which is now just a re-export of api_gateway.iam. Also drop the cache_clear() from tests/aws/api_gateway/conftest.py:11, redundant after this.

tests/aws/pulumi_mocks.py
```diff
@@ -285,6 +293,48 @@ def _fill(template: Any, subs: dict[str, str]) -> Any:
     return template
 
 
+def _add_http_api_outputs(
```
michal-stlv, line 296: One refactor ask for this file, three parts:

`_add_http_api_outputs` - half is dead. A mock resource's `.id` comes from `new_resource`'s return value, not the state dict, so the integration/route/authorizer/apiMapping branches and the Api `id`/`api_endpoint` lines never do anything - cut them all and the full aws suite stays green. Keep Api `arn` + `execution_arn`, Stage `invoke_url`, DomainName block.

The 7 `created_http_api_*` accessors - let's not grow per-service wrappers, the mocks API is consolidating on `assert_res`/`assert_res_counts`. Two are dead already (`created_http_api_authorizers`, `created_http_api_mappings` - no callers). For the rest: make `_filter_created` public as `created(typ: R, name: str | None = None)` and call that where you genuinely need all-of-type-X (coexistence, route-keys table, tagging lambdas).

Known-name call sites don't need any accessor - `assert_res` works without inputs and takes partial ones: timeout test becomes `assert_res("my-api-integration-my-api-functions-simple_handler", R.HTTP_API_INTEGRATION, {"timeoutMilliseconds": 30000.0}, partial=True)`, disable-execute becomes `assert_res("my-api", R.HTTP_API, {"disableExecuteApiEndpoint": True}, partial=True)`. Then all 7 `created_http_api_*` wrappers go.

one more thing for the branches that stay: the keys here are snake_case (`execution_arn`, `invoke_url`, `domain_name_configuration`) while the NOTE above `OUTPUT_TEMPLATES` says outputs must be camelCase and snake_case is undocumented leniency. and the type strings are raw (`"aws:apigatewayv2/api:Api"`) where the rest of the file uses the `R` enums, `new_resource` itself checks `R.HTTP_API` a few lines down.

tests/aws/cloudfront/test_cloudfront_adapter_api_gateway.py
```diff
@@ -162,31 +166,37 @@ def test_adapter_with_various_indices(idx):
...
-        assert origin_config.origin_access_controls is None
-        assert origin_config.origins["custom_origin_config"] == {
-            "http_port": 80,
-            "https_port": 443,
-            "origin_protocol_policy": "https-only",
-            "origin_ssl_protocols": ["TLSv1.2"],
-        }
-        assert origin_config.ordered_cache_behaviors["path_pattern"] == "/api/*"
-        assert origin_config.ordered_cache_behaviors["allowed_methods"] == [
+def test_router_creates_cloudfront_origin_for_rest_api(
```
michal-stlv, line 25: Now that this Router test pins the real origin behavior, the Mock-based tests in this file can go, above and below this one - they assert Mock attributes or re-implement the adapter logic inside the test and compare it with itself. We planned to delete them in a follow-up, but since this file is already touched here, no reason to merge them. Two things to carry over: add oacs == 0 to this test (only the http Router test pins that API Gateway gets no origin access control), and keep test_match_api_component - the False-for-non-API half is pinned nowhere else. The two JS-content tests (test_cloudfront_js_function_generation_for_api_gateway, test_js_function_path_lengths_for_api_gateway) are near-duplicates of the ones in the lambda/s3/url adapter files, so deleting them loses nothing.

stelvio/aws/api_gateway/rest_api/rest_api.py
```diff
@@ -165,11 +165,7 @@ def invoke_url(self) -> Output[str]:
 
     @property
     def url(self) -> Output[str]:
-        if self.domain_name is None:
-            return self.resources.stage.invoke_url
-        if self._config.base_path is None:
-            return Output.from_input(f"https://{self.domain_name}")
-        return Output.from_input(f"https://{self.domain_name}/{self._config.base_path}")
+        return self._custom_domain_url(self.resources.stage.invoke_url)
```
michal-stlv: consolidation is good but it changed behavior. the fallback arg is evaluated eagerly so api.url now triggers resource creation even when a domain is set. before this commit the domain branches returned without touching self.resources. HttpApi guards exactly this in its url property (self.resources.stage.invoke_url if domain is None else None) and http-api.md in this PR documents that with a domain url does not create resources by itself. so RestApi is now inconsistent with HttpApi and with that contract keep the consolidation, just mirror the guard:
```py
 return self._custom_domain_url(
     self.resources.stage.invoke_url if self.domain_name is None else None
 )
```
custom_domain_url fallback type becomes Output[str] | None. lazy matters here: with a domain you can pass api.url to another component and still add routes after. eager turns that into a RuntimeError for nothing.

stelvio/aws/api_gateway/iam.py
```diff
@@ -0,0 +1,69 @@
...
+        return Account(
+            "api-gateway-account",
+            cloudwatch_role_arn=role.arn,
+            opts=ResourceOptions.merge(
+                provider_opts,
+                ResourceOptions(retain_on_delete=True),
+            ),
+        )
+
+    return existing_account.cloudwatch_role_arn.apply(handle_existing_role)
```
michal-stlv, line 51: While moving we lost all the comments that I find extremely helpful as it took me 3 times to get this right :)
```suggestion
@cache
def _create_api_gateway_account_and_role() -> Output[Account]:
    provider_opts = ResourceOptions(provider=ProviderStore.aws())

    # Read existing account to check if CloudWatch role is already configured.
    # API Gateway has one Account settings per region — this reads the current state.
    existing_account = Account.get(
        "api-gateway-account-ref", "APIGatewayAccount", opts=provider_opts
    )

    def handle_existing_role(existing_arn: str) -> Account:
        if existing_arn:
            # Account already has a CloudWatch role configured (by us, CDK, or manual).
            # Leave it alone — don't try to re-adopt or override.
            logger.info("CloudWatch role already configured: %s", existing_arn)
            return existing_account

        # No role configured — create one and set it on the Account.
        logger.info("No CloudWatch role found, creating Stelvio configuration")
        role = _create_api_gateway_role(provider_opts)
        return Account(
            "api-gateway-account",
            cloudwatch_role_arn=role.arn,
            opts=ResourceOptions.merge(
                provider_opts,
                ResourceOptions(retain_on_delete=True),
            ),
        )

    return existing_account.cloudwatch_role_arn.apply(handle_existing_role)
```

README.md
```diff
@@ -65,7 +65,8 @@ def run() -> None:
 Stelvio provides high-level components for the most common AWS services:
 
 - **[Function](https://stelvio.dev/docs/components/aws/lambda/)** (AWS Lambda)
-- **[Public API](https://stelvio.dev/docs/components/aws/rest-api/)** (REST API)
+- **[REST API](https://stelvio.dev/docs/components/aws/rest-api/)** (API Gateway v1)
+- **[HTTP API](https://stelvio.dev/docs/components/aws/api_gateway/http_api/)** (API Gateway v2)
```
michal-stlv: wrong link. points at `components/aws/api_gateway/http_api/` but the page is `components/aws/http-api/`, flat next to rest-api, dash not underscore. separate thought, not this PR: once websocket api lands maybe the three api pages deserve their own path structure in docs, now everything is flat.

stelvio/aws/api_gateway/rest_api/iam.py
```diff
@@ -1,75 +1,3 @@
-import logging
-from functools import cache
+from stelvio.aws.api_gateway.iam import _create_api_gateway_account_and_role
```
michal-stlv: this file is now a dead re-export shim. the only importer left is the stale `reset_caches` import in http_api/conftest.py which goes away per my comment there. delete the whole file, no compat shims for private functions.

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
```diff
@@ -0,0 +1,759 @@
+"""Tests for HTTP API authorizers."""
```
michal-stlv, line 1: who'd say ? :)

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,740 @@
...
+
+
+def test_http_api_rejects_invalid_config_type():
+    with pytest.raises(TypeError, match="Invalid config type"):
+        HttpApi("my-api", config=123)  # type: ignore[arg-type]
+
+
+@pytest.mark.parametrize("case", HTTP_API_CASES, ids=lambda case: case.test_id)
+@pulumi.runtime.test
+def test_http_api_resource_graph(pulumi_mocks, case):
```
michal-stlv, line 317: no runtime test for `HttpApi` customize keys. `api`, `stage`, `log_group` and `api_mapping` are never exercised, only the `ApiDomain` ones are covered in the domain file. if the `_customizer("api_mapping", ...)` wiring breaks everything stays green. can you add a test that customizes all four keys and asserts the values land on the created resources? `api_mapping` needs a domain so one test with `domain_name` set covers all of them.

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,740 @@
...
+        assert len(apis) == 2
+        assert {a.name for a in apis} == {TP + "user-api", TP + "admin-api"}
+
+        # Every resource name across both APIs is unique (no cross-API collisions)
+        all_names = [r.name for r in pulumi_mocks.created_resources]
+        assert len(all_names) == len(set(all_names)), "Resource names collide across APIs"
+
+        # Per-API: 2 routes + 2 integrations + 2 functions + 2 permissions
+        for api_name in ("user-api", "admin-api"):
+            routes = [r for r in pulumi_mocks.created_http_api_routes() if api_name in r.name]
```
michal-stlv: two things in this test. the asserts filter with `api_name in r.name` and then only count, so a wrongly named resource that happens to contain `user-api` still passes. assert the exact name sets instead. and the manual wait list at the bottom is not needed, `when_http_api_ready` already takes a list of apis, using it kills the list and the comment explaining why it is built by hand.

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,740 @@
...
+    "bad_arn",
+    [
+        "not-an-arn",
+        "arn:aws:s3:::bucket",  # Wrong service
+        "arn:aws:cognito-idp:us-east-1:123:something/else",  # Wrong resource prefix
+        "arn:aws:cognito-idp::123:userpool/abc",  # Missing region
+        "arn:aws:cognito-idp:us-east-1:123:userpool/",  # Empty pool id
+    ],
+)
+def test_http_api_cognito_authorizer_invalid_user_pool_arn(bad_arn):
```
michal-stlv: this table belongs in `test_http_api_authorizers.py` with the other cognito tests. and when you move it, `test_cognito_authorizer_rejects_malformed_user_pool_arn` over there can go, it duplicates the first row of this table.

tests/aws/api_gateway/http_api/test_http_api_route.py
```diff
@@ -0,0 +1,146 @@
...
+        assert functions[0].name == "test-test-users-function"
+
+        permissions = pulumi_mocks.created_permissions()
+        assert len(permissions) == 1
+        assert permissions[0].inputs["function"] == "test-test-users-function-test-name"
+
+    when_http_api_ready(api, check)
+
+
+@pytest.mark.parametrize(
```
michal-stlv: three untested branches around `Function` handlers. `route(..., fn, memory=512)` should raise "Cannot combine a Function handler with function options" but nothing pins it, the same guard in `add_lambda_authorizer` is untested too, and `add_lambda_authorizer` with a `Function` instance has no happy path test at all. i deleted all three branches and the whole http_api suite stayed green. one more row in this table plus two authorizer tests would cover it.

tests/aws/api_gateway/http_api/test_http_api_route.py
```diff
@@ -0,0 +1,146 @@
+import pulumi
```
michal-stlv: this file and `test_http_api_routes.py` are one letter apart and both test `route()`. merge them into `test_http_api_routes.py`? together they are about 200 lines, validation tables first, handler config tests after, and the confusing pair of names goes away.

tests/aws/api_gateway/http_api/test_http_api_domain.py
```diff
@@ -0,0 +1,335 @@
...
+    ApiDomain("my-api-domain", domain_name="other.example.com")
+    api = HttpApi("my-api", domain_name="api.example.com")
+    api.route("GET", "/users", "functions/simple.handler")
+
+    with pytest.raises(ValueError, match="Duplicate Stelvio component name"):
+        _ = api.resources
+
+
+@pulumi.runtime.test
+def test_http_api_domain_duplicate_mapping_key_raises(pulumi_mocks, app_context_with_dns):
```
michal-stlv, line 289: this only covers duplicates with explicit keys. two apis on the same domain with no mapping key at all should also raise, that is the `(root)` branch in `register_mapping`. worth a second case here, if that check ever special cases `None` the tests stay green but the deploy fails with two root mappings.

tests/aws/api_gateway/http_api/test_http_api_config.py
```diff
@@ -0,0 +1,6 @@
+from stelvio.aws.api_gateway.http_api import HttpApiConfig, HttpApiConfigDict
+from tests.test_utils import assert_config_dict_matches_dataclass
+
+
+def test_http_api_config_dict_matches_http_api_config():
```
michal-stlv: this file is down to one contract test, fold it into `test_http_api.py` and delete the file. that is where other components keep theirs, `test_user_pool.py`, `test_queue.py` and `test_dynamodb_table.py` all have the same test inline.

tests/aws/pulumi_mocks.py
```diff
@@ -293,6 +343,8 @@ def __init__(self):
         self.created_resources: list[MockResourceArgs] = []
 
     def new_resource(self, args: MockResourceArgs) -> tuple[str, dict[str, Any]]:
+        if args.typ == R.CERTIFICATE_VALIDATION:
+            args.inputs["validationRecordFqdns"] = ["_test.api.example.com"]
```
michal-stlv: this overwrite makes the domain test assert circular. the component passes `[validation_record.name]` (acm.py:107), the mock replaces it with a constant, and `assert_api_domain_graph` then asserts that same constant. wrong wiring in prod stays green, and any test using a domain other than api.example.com would fail. the actual gap is that the generic `name` output is the `-test-name` string instead of the input name. add `"name": "{in[name]}"` to the `CLOUDFLARE_RECORD` and `ROUTE53_RECORD` templates, that wins over the generic rule, then this whole `if` can go. i tried it, all 1596 aws tests pass.

tests/aws/cloudfront/test_cloudfront_adapter_api_gateway.py
```diff
@@ -162,31 +166,37 @@ def test_adapter_with_various_indices(idx):
...
+                    "originProtocolPolicy": "https-only",
+                    "originSslProtocols": ["TLSv1.2"],
+                },
+                "domainName": (f"{tid(TP + 'edge-api')}.execute-api.us-east-1.amazonaws.com"),
+                "originId": tid(TP + "edge-api"),
+                "originPath": "/v1",
+            }
+        ]
+
+        behavior = distribution.inputs["orderedCacheBehaviors"][0]
```
michal-stlv: the behavior asserts stop short. `functionAssociations` is never checked, so the rewrite function can exist but never be attached to the behavior and this stays green, that wiring was pinned by the deleted mock test. compress, viewerProtocolPolicy, targetOriginId and forwardedValues are not checked either. make it one exact dict == like the origins assert above, with the rewrite function arn in functionAssociations.

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
```diff
@@ -0,0 +1,131 @@
...
+        origins = distribution.inputs["origins"]
+        assert len(origins) == 1
+        origin = origins[0]
+        assert origin["domainName"] == (
+            f"{tid(TP + 'edge-api')[:8]}.execute-api.us-east-1.amazonaws.com"
+        )
+        assert origin.get("originPath") is None
+        assert origin["customOriginConfig"]["originProtocolPolicy"] == "https-only"
+
+        behaviors = distribution.inputs["orderedCacheBehaviors"]
```
michal-stlv: same ask as in the rest adapter file: assert the whole behavior as one exact dict == including `functionAssociations` with the rewrite function arn, right now the function is only checked to exist by name so dropping the association stays green. the origins asserts can be one exact list == too, like the rest twin does. and the rewrite code is unasserted here, the rest test pins it with `strip_path_pattern_function_js("/api")`.

tests/integration/test_http_api.py
```diff
@@ -0,0 +1,319 @@
...
+        export_function(fn)
+        export_http_api(api)
+
+    outputs = stelvio_env.deploy(infra)
+
+    assert_http_api_routes(
+        outputs["http_api_sharedhttpapi_id"],
+        expected_route_keys={"GET /one", "POST /two"},
+    )
+    assert "function_sharedhttp_arn" in outputs
```
michal-stlv: three things across this file. hitting both routes proves they work but not that they share the exported function, two separate lambdas from the same handler would return the same 200s. get the api integrations via boto3 and assert both `IntegrationUri`s equal the `function_sharedhttp_arn` output, that was the original ask here. the jwt and cognito tests only check `AuthorizerType` and the route auth type, issuer and audiences are never verified and the cognito test exports the pool without using the output, assert the authorizer config. and url joining is inconsistent, four places do `f"{base_url}/path"` and two do `rstrip("/") +`, normalize once, simplest is stripping in `export_http_api`.

tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -0,0 +1,740 @@
...
+    tn,
+)
+from .conftest import TP, when_http_api_ready
+
+pytestmark = pytest.mark.usefixtures("project_cwd")
+
+# --- Shared constants / templates -------------------------------------------
+# The PulumiTestMocks derive the API id from the resource id: api_id = tid(name)[:8].
+HTTP_API_ID = tid(TP + "my-api")[:8]
+API_EXECUTION_ARN = f"arn:aws:execute-api:{DEFAULT_REGION}:{ACCOUNT_ID}:{HTTP_API_ID}"
```
michal-stlv: small ones bundled so they are not ten threads:

- `API_EXECUTION_ARN` and `ROUTE_PERMISSION_SOURCE_ARN` here are dead, and `HTTP_API_ID` plus the invoke arn template are copy pasted in the authorizers file, hoist the shared ones into the package conftest
- this import block and the same one in the authorizers file only stay multi line because of the trailing comma, it fits on one line, the domain file already does that
- new files use `import pytest` with prefixed calls, we are migrating to `from pytest import mark, param, raises` style (see `test_vpc.py`), new files are the cheapest place to start with it
- `raises(TypeError)` at authorizers L501 has no match, use match="UserPoolClient audiences require"
- the duplicate authorizer name test only goes through `add_jwt_authorizer`, parametrize over all three add methods and the other two get covered for free
- the conflicting config params in `test_http_api_route.py` are 3 and 4 tuples with a len guard at unpack, make them all 4 tuples
- `test_tagging_contract.py`: `_trigger_api_custom_domain`, `_build_email` and `_trigger_email` are dead, the TagCase rows use identical inline lambdas, delete the helpers or point the rows at them
- `test_customization_sync.py`: drop `"authorizers"` from the HttpApi exclude set, it is not a `HttpApiResources` field so today it does nothing and later it would swallow the drift alert. also drop `"dns_record"` from both ApiDomain exclude sets, field and key match by name so nothing needs excluding
- authorizer ttl validation is `0 <= ttl <= 3600` but only 0 and 3601 have rows, add -1 and 3600
- `test_dns_http_api.py` hardcodes us-east-1 in the expected domain, use `stelvio_env.aws_region`

michal-stlv (changes requested):
tests came out really good, the spec case rewrite is exactly what i hoped for and everything from the first round is addressed. this round is mostly quality threads on the new tests, a few real gaps among them (the mock overwrite of validationRecordFqdns, functionAssociations never asserted in the two Router tests, the untested Function handler branches) plus a couple of things in the prod files, the url guard on RestApi being the one i care most about. i have NOT reviewed the docs yet, that I will do in separate review - next

docs/docs/components/aws/http-api.md
```diff
@@ -0,0 +1,495 @@
...
+This guide explains how to build HTTP APIs with Stelvio using
+[Amazon API Gateway HTTP APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api.html).
+You'll learn how to define routes, protect them with authorizers, configure CORS,
+and put your API behind a custom domain.
+
+HTTP APIs are the second generation of API Gateway. They're cheaper and faster than
+REST APIs, use the Lambda payload format 2.0, and come with native CORS support and
+auto-deployed stages. Use `HttpApi` for new Lambda-backed HTTP endpoints unless you
+need behavior from API Gateway REST APIs, such as edge-optimized endpoints, the
+REST API Lambda event shape, or Stelvio's REST API CORS response helpers. For
```
michal-stlv: listing the cors response helpers as a reason to pick `RestApi` reads backwards. the helpers exist because rest api cannot do cors at the gateway, they are a workaround for that, not a feature you would choose it for. i'd drop that item from this list, from the closing section, and from the same sentence in the rest-api.md intro. the event shape item is close behind: nobody needs the v1 format for its own sake, it only matters when existing handlers or libraries already expect v1 events. reword it to say that, something like "when your existing handlers or libraries expect the v1 event format". that leaves edge-optimized endpoints and token and request authorizers as the real feature reasons.

docs/docs/components/aws/http-api.md
```diff
@@ -0,0 +1,495 @@
...
+# API with a named stage instead of the default $default stage
+api = HttpApi("users-api", stage_name="production")
+```
+
+Available configuration options:
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| `domain_name` | `None` | Custom domain name for an API-owned domain. Requires a DNS provider. |
+| `domain` | `None` | Shared `ApiDomain` component for mapping multiple HTTP APIs to one domain. |
```
michal-stlv: worth noting here that `domain_name` and `domain` can't be combined, the code raises when both are set.

docs/docs/components/aws/http-api.md
```diff
@@ -0,0 +1,495 @@
...
+| `api_mapping_key` | `None` | Path segment for a custom domain mapping, such as `admin` or `partners/v1`. Requires `domain_name` or `domain`. |
+| `stage_name` | `"$default"` | HTTP API stage name. Use `"$default"` or a name containing letters, numbers, hyphens, and underscores. The `$default` stage serves at the root of the API URL; a named stage adds a path segment, such as `/production`. |
+| `cors` | `None` | CORS configuration. Use `True` for permissive defaults or `CorsConfig` for explicit settings. |
+| `disable_execute_api_endpoint` | `False` | Disable the default `execute-api` hostname. Requires a custom domain. |
+| `access_log_retention_days` | `30` | CloudWatch access log retention in days. Set to `"forever"` to keep logs indefinitely. |
```
michal-stlv: "retention in days" reads like any number works, but only CloudWatch's fixed values are valid, `access_log_retention_days=45` raises. maybe say "one of CloudWatch's retention values (1, 3, 5, 7, 14, 30, 60, 90, ...)" so nobody finds out at deploy time.

docs/docs/components/aws/http-api.md
```diff
@@ -0,0 +1,495 @@
...
+
+`{proxy+}` is a greedy path parameter: it matches the remaining path segments and
+must be the final segment. For example, `/files/{proxy+}` matches
+`/files/images/avatar.png`, but not `/files` itself.
+
+HTTP API also supports a special `$default` route that catches any request that
+doesn't match another route:
+
+```python
+api.route("ANY", "$default", "functions/fallback.handler")
```
michal-stlv, line 123: worth a sentence here that `$default` only works with `"ANY"`, any other method raises. the example uses ANY so it works, but the rule is invisible.

docs/docs/components/aws/http-api.md
```diff
@@ -0,0 +1,495 @@
...
+```
+
+If you'd rather return an IAM policy response, set `simple_response=False`.
+
+**Configuration options:**
+
+- `name`: Unique authorizer name within the API
+- `handler`: Lambda function path, config, or `Function` instance
+- `identity_sources`: List of selection expressions to extract identity from (required)
+- `ttl`: Cache TTL in seconds (default: 300). Set to `0` to disable authorizer caching.
```
michal-stlv: two small ones for this section: `ttl` is validated to 0-3600, the bullet only says 0 disables caching, add the upper bound. and in the cognito authorizer, mixing forms is not free: `UserPoolClient` audiences require passing the `UserPool` component, an ARN string plus client components raises `TypeError`, and the clients must belong to that pool. one sentence under the cognito example saves the trial and error.

docs/docs/components/aws/http-api.md
```diff
@@ -0,0 +1,495 @@
...
+```python
+api = HttpApi("public-api", cors=True)
+```
+
+With `cors=True`, Stelvio allows all origins (`"*"`), all standard methods, and
+all headers, with credentials disabled.
+
+For production, configure origins, methods, and headers explicitly. `CorsConfig`
+accepts either strings or lists for origins, methods, and headers; HTTP APIs
+normalize them to API Gateway v2 lists.
```
michal-stlv: one gotcha worth stating: `"*"` has to be the plain string, `allow_origins=["*"]` raises. the "strings or lists" phrasing invites putting the wildcard in a list.

docs/docs/components/aws/http-api.md
```diff
@@ -0,0 +1,495 @@
...
+    cors=CorsConfig(
+        allow_origins=[
+            "https://app.example.com",
+            "https://admin.example.com",
+        ],
+        allow_methods=["GET", "POST", "PATCH", "DELETE"],
+        allow_headers=["content-type", "authorization"],
+        allow_credentials=True,
+        max_age=3600,
+    ),
```
michal-stlv, line 369: `CorsConfig` also has `expose_headers`, the example and text never mention it. worth one line, it is the field people need when the browser can't read a custom response header.

docs/docs/components/aws/http-api.md
```diff
@@ -0,0 +1,495 @@
...
+| `HttpApi` | `api` | [ApiArgs](https://www.pulumi.com/registry/packages/aws/api-docs/apigatewayv2/api/#inputs) | The API Gateway v2 HTTP API. |
+| `HttpApi` | `stage` | [StageArgs](https://www.pulumi.com/registry/packages/aws/api-docs/apigatewayv2/stage/#inputs) | The auto-deploy stage. |
+| `HttpApi` | `log_group` | [LogGroupArgs](https://www.pulumi.com/registry/packages/aws/api-docs/cloudwatch/loggroup/#inputs) | The CloudWatch access log group. |
+| `HttpApi` | `api_mapping` | [ApiMappingArgs](https://www.pulumi.com/registry/packages/aws/api-docs/apigatewayv2/apimapping/#inputs) | The custom domain mapping when `domain_name` or `domain` is set. |
+| `ApiDomain` | `certificate` | [CertificateArgs](https://www.pulumi.com/registry/packages/aws/api-docs/acm/certificate/#inputs) | The ACM certificate for the custom domain. |
+| `ApiDomain` | `domain` | [DomainNameArgs](https://www.pulumi.com/registry/packages/aws/api-docs/apigatewayv2/domainname/#inputs) | The API Gateway v2 custom domain. |
+| `ApiDomain` | `dns_record` | DNS provider record args | The DNS record pointing the custom domain to API Gateway. |
+
+
+## REST API or HTTP API?
```
michal-stlv: this section repeats the intro almost word for word. keep one, i'd keep the intro since that is where people decide, and drop this section or reduce it to a link back to the rest api guide.

docs/changelog.md
```diff
@@ -2,9 +2,13 @@
 
 ## 0.10.0b6 (2026-MM-DD)
 
-### Api Gateway changes (will be changed in merge queue)
-**This is a placeholder for changes to API Gateway components.**
-This PR adds the `HttpApi` component.
+### Breaking Changes
+- `Api` component (AWS API Gateway v1) is renamed to `RestApi`.
+→ [REST API Guide](docs/components/aws/rest-api.md)
```
michal-stlv, line 8: two rendering nits here: the arrow link lines sit right under the bullet with no blank line, so they render inside the bullet, the older entries (vpc, cognito) have them as their own paragraph. and further down, the blank line between the vpc guide link and the component customization heading got deleted, it still renders but it is the only heading in the file without one, looks accidental.

michal-stlv (commented):
docs round, as promised in the first review. the http api guide is genuinely good, it follows the shape of our other component guides, and i verified every factual claim against the source, all check out, including the `api.url` note which is accurate for `HttpApi`. the 9 comments are all small: a few validation rules the code enforces but the docs don't state, one argument in the rest-vs-http comparison that reads backwards, and two changelog formatting nits. nothing blocking beyond what round 1 already asks.

tests/aws/api_gateway/http_api/conftest.py
sebst: autouse fixture in `tests/aws/conftest.py`

tests/aws/pulumi_mocks.py
sebst: `created_http_api_*` wrappers removed

tests/aws/cloudfront/test_cloudfront_adapter_api_gateway.py
sebst: dupes deleted; kept Router + `test_match_api_component`; added `oacs == 0`

stelvio/aws/api_gateway/rest_api/rest_api.py
sebst: `invoke_url if domain_name is None else None`

stelvio/aws/api_gateway/iam.py
sebst: restored

README.md
sebst: fixed

stelvio/aws/api_gateway/rest_api/iam.py
sebst: deleted

tests/aws/api_gateway/http_api/test_http_api_authorizers.py
sebst: :)

tests/aws/api_gateway/http_api/test_http_api.py
sebst: `test_http_api_customize_applies_to_resources`

tests/aws/api_gateway/http_api/test_http_api.py
sebst: asserts exact name sets; uses `when_http_api_ready([api1, api2], ...)`

tests/aws/api_gateway/http_api/test_http_api.py
sebst: moved to authorizers

tests/aws/api_gateway/http_api/test_http_api_route.py
sebst: done

tests/aws/api_gateway/http_api/test_http_api_route.py
sebst: merged; `test_http_api_route.py` deleted

tests/aws/api_gateway/http_api/test_http_api_domain.py
sebst: `test_http_api_domain_duplicate_root_mapping_raises`

tests/aws/api_gateway/http_api/test_http_api_config.py
sebst: `test_http_api_config.py` deleted

tests/aws/pulumi_mocks.py
sebst: overwrite removed

tests/aws/cloudfront/test_cloudfront_adapter_api_gateway.py
sebst: fixed

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
sebst: exact origins/behavior (incl. associations) + `strip_path_pattern_function_js("/api")`

tests/integration/test_http_api.py
sebst: shared IntegrationUri assert; JWT/Cognito issuer+audiences; `export_http_api` strips trailing `/`

tests/aws/api_gateway/http_api/test_http_api.py
sebst: done

docs/docs/components/aws/http-api.md
sebst: dropped CORS helpers; reworded to existing handlers/libraries expecting v1

docs/docs/components/aws/http-api.md
sebst: table + note that combining them raises

docs/docs/components/aws/http-api.md
sebst: done

docs/docs/components/aws/http-api.md
sebst: added

docs/docs/components/aws/http-api.md
sebst: done

docs/docs/components/aws/http-api.md
sebst: done

docs/docs/components/aws/http-api.md
sebst: done

docs/docs/components/aws/http-api.md
sebst: section removed; Next Steps points at RestApi guide

docs/changelog.md
sebst: done

stelvio/aws/api_gateway/domain.py
```diff
@@ -20,6 +20,7 @@
 class ApiDomainResources:
     acm_domain: AcmValidatedDomain
     custom_domain: pulumi_aws.apigatewayv2.DomainName
+    dns_record: Record
```
michal-stlv, line 23: not about this line, it is the only one i can anchor to in this file. `ApiDomain` is planned to be shared beyond http api (#215, one v2 domain for http + rest + websockets), so if it is reusable it should not live under `http_api/`. moving it after release breaks anyone importing from the `http_api` path, moving it now is a rename plus a few import lines since everything re-exports through `stelvio.aws.api_gateway` anyway. not a blocker, and it can also be folded into #215, but before merge is the cheap moment.

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
```diff
@@ -0,0 +1,143 @@
...
+                    "httpsPort": 443,
+                    "originProtocolPolicy": "https-only",
+                    "originSslProtocols": ["TLSv1.2"],
+                },
+                "domainName": f"{api_id}.execute-api.us-east-1.amazonaws.com",
+                "originId": api_id,
+            }
+        ]
+
+        behavior = distribution.inputs["orderedCacheBehaviors"][0]
```
michal-stlv: small one on the new exact asserts, applies here and in the rest adapter file: the behavior is read as `orderedCacheBehaviors[0]`, so a stray second behavior would pass unnoticed, the old version had `len(behaviors) == 1`. simplest is to assert the whole list like origins already does, `assert distribution.inputs["orderedCacheBehaviors"] == [{...}]`, plus one `len(distributions) == 1` since the distribution is indexed the same way. not a blocker.

michal-stlv (commented):
thanks for the quick turnaround. three small ones left.

the third one has no line to hang on, it comes from the stack branch: `VALID_LOG_RETENTION_DAYS` in `stelvio/aws/api_gateway/validators.py` is missing five values cloudwatch supports (1096, 2192, 2557, 2922, 3288), so `access_log_retention_days=1096` raises even though aws takes it, and the docs line now says "one of CloudWatch's retention values". one line there plus a row in the validation test.

tests/aws/cloudfront/test_cloudfront_adapter_http_api.py
sebst: changed

stelvio/aws/api_gateway/domain.py
sebst: would really leave that to follow up pr as discussed

michal-stlv (approved):

PR #239, opened 2026-07-26
test: rewrite rich deployment handler tests as behavioral output tests

michal-stlv:
Rewrote the rich deployment handler tests to be behavioral: feed real Pulumi events in,
assert exactly what the user sees (terminal render, --json, --stream, completion). Old
tests asserted internals — they broke on refactors but missed real output bugs. Every
test is mutation-checked now.

Along the way this surfaced a bunch of real bugs in the output code (wrong drift values
on refresh, leaked Pulumi internals like the unknown-value uuid, missing resource names
in stream errors, wrong JSON for failed orphans, blank line issues...) — fixed here,
plus some dead code removed.

Manually retested the whole thing against real AWS: deploy, diff, refresh, failed
deploy, recovery, destroy — all three output formats.


tests/test_rich_deployment_handler.py
```diff
@@ -1,10 +1,17 @@
-"""Tests for RichDeploymentHandler (Phase 1 + Phase 2 CLI redesign)."""
+"""Behavioral tests for RichDeploymentHandler: events in, the four public outputs out.
+
+Every test feeds Pulumi engine events through ``handle_event`` and asserts on what the
+user sees — the rendered terminal frame, the ``--json`` payload, the ``--stream`` events,
+or the completion frame. See notes/rich-deployment-behavioral-tests.md for the rules.
```
sebst: which notes? ;)

tests/test_rich_deployment_handler.py
```diff
@@ -170,1765 +166,1201 @@ def _failed_event(urn: str, resource_type: str, timestamp: int = 1001) -> Engine
...
-            }
-        },
-        new_inputs={
-            "environment": {
-                "variables": {"STLV_USERS_TABLE_ARN": "04da6b54-80e4-46f7-96ec-b56ff0331ba9"}
-            }
-        },
-    )
-
-    lines = format_property_diff_lines(r, line_width=60)
```
sebst, line 1508: the 60 line width has been dropped from the tests?

sebst (dismissed):

tests/test_rich_deployment_handler.py
michal-stlv: removed

tests/test_rich_deployment_handler.py
michal-stlv: The old test called the formatter directly, so the 60 never came from a real console. Truncation is now pinned through real rendered frames — narrow at width 80 (test_preview_truncates_long_values_in_the_middle_on_narrow_terminal + add/delete budget tests), wide at 200 — and the output<string> marker is pinned on both render and --json. As of 1ad9246 the unknown-output sentinel is replaced at ingestion, so truncation can't garble it at any width.

sebst (approved):

PR #240, opened 2026-07-28
docs: contributor guides for components, tests and documentation

michal-stlv:
First pass of the contributing docs we discussed on Monday.

Four pages that will publish under stelvio.dev/docs/contributing/: writing
components, unit tests, integration tests and documentation. Each one is built
around real code from the repo as role models (cron/topic/vpc for components,
test_vpc for unit tests, vpc.md/queues.md for docs) rather than abstract rules.

CONTRIBUTING.md now points to them and its tests section is fixed: integration
tests are the release gate, so infra changes should add or update them — running
them is the optional part if you don't have an account handy, just say so in the
PR. The stale tier commands are gone too; run_all.sh is now the single source of
truth for tiers and worker counts.

These pages describe the current and desired state of the codebase, and all PRs
are expected to follow them — implementation, tests, docs. That goes for both of
us as much as for any new contributor. They're also meant to be living docs: we
should keep updating them as we go, based on PR comments and whatever we learn.

Ideally I'd like to merge this as is and handle any updates in follow-up PRs,
but happy to discuss changes here if needed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,15 @@ Key areas of the codebase:
 
 ## Development
 
+### Writing code and docs
+
+The contributor guides on stelvio.dev cover the patterns we follow:
+
+- [Writing components](https://stelvio.dev/docs/contributing/components/)
+- [Writing unit tests](https://stelvio.dev/docs/contributing/unit-tests/)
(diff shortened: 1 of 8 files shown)


sebst (approved):
> Ideally I'd like to merge this as is and handle any updates in follow-up PRs,
but happy to discuss changes here if needed.

Agreed

PR #244, opened 2026-07-31
fix: proper parenting for DNS records

### LINKED ISSUE #209: Improve usage of `create_caa_record`

sebst:
closes #209


stelvio/dns.py
```diff
@@ -17,22 +17,42 @@ def pulumi_resource(self) -> Resource:
 
 
 class Dns(Protocol):
```
michal-stlv, line 17: Since this PR already breaks the `Dns` protocol, can we finish the DNS layer here and break it only once? Then the docs and changelog also get written once. Four things, all small.

1. One method. `create_caa_record` is `create_record` with `content` instead of `value`. Its only caller is the ACM validation record in `acm.py`, a CNAME, not a CAA record. Can we drop it and call `create_record` with `value=`? That is #209. A real CAA record would be a new method anyway, on Cloudflare it needs `data` with flags, tag and value, not `content`.

2. `name` on `Record`. The docs example returns `Record(record)`, but `acm.py` line 112 reads `validation_record.name` and the base `Record` has no `name`. An adapter written from the docs fails on the first custom domain deploy. Can we make `Record` a frozen dataclass with two fields, `pulumi_resource` and `name`? That is all Stelvio reads from a record. `name` is explicit because `pulumi.Resource` has no `name` and some providers store names relative to the zone. Providers return `Record(record, record.name)` and the two `*PulumiResourceAdapter` classes can go. Their `type` and `value` have no readers in `stelvio/`, every `Resources` field is typed as the base `Record`, and the docs never mention them. The only `value` reader is the integration export helper for DKIM and DMARC, and it can read the Route 53 resource directly, like it does for CloudFront. Or was there a plan for `type` and `value`?

3. `Dns` as a plain protocol. Every caller checks for a missing provider first and raises `DnsProviderNotConfiguredError`, so the `raise NotImplementedError` bodies in `Dns` only run when a subclass forgot the method, and then the message is wrong. They never run for an adapter written from the docs, which has no base class. An ABC would only guard a one-method class that forgot its one method. Can we make the bodies `...` and drop the `dns.Dns` base from `Route53Dns`, `CloudflareDns` and `MockDns`? Then nothing inherits from `Dns`, a missing method is a plain `AttributeError`, and `Dns` is only the shape, like `Home` and `Permission`.

4. Types. `name` and `record_type` are `str`, but `acm.py` passes Outputs for both. `Input[str]`, like `value`.

After this a custom adapter is one class with one method that returns `Record(record, record.name)`. Docs: one method, same example, and the "Breaking change" box can go, the paragraph above it already says it. Changelog note is on the line.

stelvio/dns.py
```diff
@@ -17,22 +17,42 @@ def pulumi_resource(self) -> Resource:
...
+        record_type: str,
+        value: Input[str],
+        ttl: int = 1,
+        *,
+        opts: ResourceOptions | None = None,
     ) -> Record:
         """
         Create a DNS record with the given name, type, and value.
+
+        Providers SHOULD forward ``opts`` to the underlying Pulumi record
```
michal-stlv: Small one. `SHOULD` in capitals reads like an RFC, and it is the only place in the package that does it. Also every Pulumi resource takes `opts`, so there is no case where a provider cannot forward it. Can it just say "Forward `opts` to the Pulumi record so Stelvio can parent it under the component"? Same for "when you can" in the docs.

stelvio/aws/acm.py
```diff
@@ -99,6 +99,7 @@ def _create_resources(self) -> AcmValidatedDomainResources:
                     "ttl": 1,
                 },
             ),
+            opts=self._resource_opts(),
```
michal-stlv, line 102: With `parent=self` the Route 53 record moves from Pulumi's default AWS provider to the component's explicit one. Same account and region, so the record does not change. Did you run `stlv diff` on an existing stack with a Route 53 custom domain? The integration tests use fresh stacks, so they would not show a replace.

stelvio/aws/api_gateway/rest_api/rest_api.py
```diff
@@ -77,6 +77,8 @@ class RestApiResources:
     log_group: cloudwatch.LogGroup
     custom_domain: DomainName | None = None
     base_path_mapping: BasePathMapping | None = None
+    dns_record: Record | None = None
```
michal-stlv: The only readers of these two fields are the new tests. We keep `Resources` for what a user customizes or reads, not for test handles, and the sync test only passes because it excludes these two. The ACM test can take the component from `ComponentRegistry.instances_of(AcmValidatedDomain)`, and `MockDns` can keep the `Record` it returns, so the record test reads the URN from there. Can we drop the two fields and the four-tuple?

tests/dns/test_dns.py
```diff
@@ -0,0 +1,47 @@
...
+
+@pytest.fixture
+def pulumi_mocks():
+    mocks = PulumiTestMocks()
+    set_mocks(mocks)
+    return mocks
+
+
+@pulumi.runtime.test
+def test_dns_adapter_without_opts_raises_type_error(pulumi_mocks):
```
michal-stlv: Can this move into `tests/aws/test_dns_providers.py`? There `app_context_with_dns` sets the context and cleans up, and it takes the `mock_dns` fixture, so a module-level `mock_dns` fixture returning `LegacyDns` gives it the adapter. Then the local `pulumi_mocks` fixture, the try/finally and the registry clearing can go, the autouse fixtures in `tests/conftest.py` already do that. `LegacyDns` can sit at module level like the other helpers there.

tests/aws/test_dns_providers.py
```diff
@@ -0,0 +1,145 @@
...
+
+class _DnsParent(Component[_DnsParentResources, dict]):
+    def __init__(self, name: str):
+        super().__init__(ProviderStore.aws(), "stelvio:test:DnsParent", name)
+
+    def _create_resources(self) -> _DnsParentResources:
+        return _DnsParentResources()
+
+
+def _assert_parented(record: Record, parent_token: str):
```
michal-stlv: `parent_token in urn` passes for any child type and name. The repo pins the whole URN, for example `test_topic.py` line 616. Under mocks the URN here is `urn:pulumi:stack::project::stelvio:test:DnsParent$aws:route53/record:Record::route53-alias`. Can the helper take the child type and name and compare the full string? The eight component tests use the same substring check, a shared helper in `tests/aws/conftest.py` would serve all of them.

tests/aws/test_dns_providers.py
```diff
@@ -0,0 +1,145 @@
...
+            "route53-alias",
+            R.ROUTE53_RECORD,
+            {
+                "name": "api.example.com",
+                "type": "CNAME",
+                "records": ["target.example.com"],
+                "ttl": 300,
+                "zoneId": "Z1234567890",
+            },
+            partial=True,
```
michal-stlv: The dict lists every input the record gets, so `partial=True` only hides a new key. Can we drop it here and on line 113?

tests/aws/test_dns_providers.py
```diff
@@ -0,0 +1,145 @@
...
+        name="_test.api.example.com",
+        record_type="CNAME",
+        content="test-validation.api.example.com",
+        opts=parent._resource_opts(),
+    )
+    return _assert_parented(record, "::stelvio:test:DnsParent$")
+
+
+@pulumi.runtime.test
+def test_mock_dns_create_record_forwards_opts_parent(pulumi_mocks):
```
michal-stlv: This tests `MockDns`, which is a test helper. If the mock stopped forwarding `opts`, the parenting tests on the components would fail already. Can we drop it?

tests/aws/api_gateway/http_api/test_http_api_domain.py
```diff
@@ -205,3 +205,27 @@ def check(_):
         )
 
     when_http_api_ready([api1, api2], check)
+
+
+@pulumi.runtime.test
+def test_api_domain_dns_records_parented(pulumi_mocks, app_context_with_dns):
```
michal-stlv: This builds a bare `ApiDomain`, so it belongs in `tests/aws/api_gateway/test_api_domain.py` with the other `ApiDomain` tests. And inside the apply callback only the three awaited URNs are settled, the graph and count asserts read resources that may not be recorded yet, and `test_api_domain.py` already checks both. Can we move it and keep only the three URN checks?

tests/aws/cognito/test_user_pool_domain.py
```diff
@@ -235,6 +235,30 @@ def check(_):
     ).apply(check)
 
 
+@pulumi.runtime.test
+def test_custom_domain_dns_record_parented(pulumi_mocks, app_context_with_dns):
```
michal-stlv: The record test and the ACM test build the same `UserPool` twice. Can they be one test with `pulumi.Output.all` of both URNs, like the acm and email tests in this PR? Same for the pairs in `test_custom_domain.py`, `test_appsync_custom_domain.py`, `test_cloudfront_router_pulumi.py` and `test_cloudfront.py`.

tests/aws/email/test_email.py
```diff
@@ -622,3 +622,29 @@ def test_email_dns_false_skips_dkim_dmarc_verification(pulumi_mocks):
     assert resources.dkim_records is None
     assert resources.dmarc_record is None
     assert resources.verification is None
+
+
+@pulumi.runtime.test
+def test_email_dns_records_parented(pulumi_mocks):
+    from ..pulumi_mocks import MockDns
```
michal-stlv: Line 11 already imports from `..pulumi_mocks`, `MockDns` can go there.

docs/docs/changelog.md
```diff
@@ -4,10 +4,32 @@
 
 ### Bug Fixes
 
+- **DNS record parenting.** DNS records created by Stelvio components (ACM validation, `ApiDomain` / `RestApi` custom-domain aliases, Email DKIM/DMARC, CloudFront/Router aliases, Cognito custom domain, AppSync custom domain) are parented under their owning component when using built-in Route 53 or Cloudflare providers. Operators may see Pulumi URN parent changes for those records (a graph nesting fix, not a DNS payload change) — Stelvio’s root-stack aliases migrate existing resources without delete/recreate where possible.
```
michal-stlv: Can we merge these two into one short bullet, like the subscription one? Something like:

**DNS record and certificate parenting.** DNS records and the `AcmValidatedDomain` created by components are now parented under the owning component. Root-stack aliases migrate existing stacks in place, no replacements.

docs/docs/changelog.md
```diff
@@ -4,10 +4,32 @@
...
 - **Default AWS region resolution.** With no `region` in `@app.config`, Stelvio now resolves the region once via the standard AWS chain (`AWS_REGION`, `AWS_DEFAULT_REGION`, profile config) and uses it everywhere. Previously a missing region put a literal `None` into region-based values, breaking `IdentityPool` deploys and AppSync DynamoDB data sources — and setting only `AWS_REGION` (which boto3 ignores) didn't work. When no region can be resolved at all, Stelvio fails fast with a clear error.
 
 ### Breaking Changes
+- **Custom `Dns` adapters must accept `opts`.** `create_record` and `create_caa_record` now require keyword-only `opts: ResourceOptions | None = None` on the public `Dns` protocol. Stelvio always passes `opts=` when creating records. Adapters that omit the parameter raise `TypeError`. There is no compatibility shim or deprecation window.
```
michal-stlv: `Dns` adapters: the protocol is now a single `create_record` that returns `Record(resource, name)` and must accept a keyword-only `opts` parameter. See [Custom DNS adapters](concepts/dns.md#custom-dns-adapters).

michal-stlv (changes requested):
great stuff,  i didn't run integration tetss my side nor idd i tested it manually in my aws accuont assuming you did that and it works

the bigger ones:
1. the comment on `dns.py`: finish the dns layer in this pr. one method, `Record` with `name`, plain protocol, `Input[str]`.
2. `rest_api.py`: the two new `Resources` fields exist only for the tests, drop them.
3. one `stlv diff` on an existing stack with a route 53 custom domain, to confirm no replacements.
4. tests: one exact-urn helper for all parenting tests, merge the record and acm pairs, move the two misplaced tests.

the rest are small wording things on the docs and changelog.

michal-stlv (approved):

PR #245, opened 2026-08-01
chore: align existing code with contributing conventions

michal-stlv:
Small sweep bringing existing code in line with the contributor guidelines (#240):

- add the missing `@final` on `Cron` — the only component without it
- `Topic`: pass `pulumi_suffix_length=0` to `safe_name` — we set an explicit `name`,
  so Pulumi never appends a suffix; no point reserving 8 chars for it
- drop dead `cleanup_state()` (zero callers). Keeping the empty state file after a
  full destroy is deliberate — same default as SST; an opt-in full purge could be a
  future feature
- test guards: `NatConfig`/`NatConfigDict` sync check in the VPC tests, and `Vpc`
  added to the keyword-only-constructor-params suite

No behavior changes.

### DIFF
--- a/stelvio/aws/topic.py
+++ b/stelvio/aws/topic.py
@@ -288,7 +288,9 @@ class Topic(Component[TopicResources, TopicCustomizationDict], LinkableMixin):
     def _create_resources(self) -> TopicResources:
         suffix = FIFO_SUFFIX if self._fifo else ""
         name = self.name.removesuffix(suffix)
-        topic_name = safe_name(context().prefix(), name, MAX_TOPIC_NAME_LENGTH, suffix=suffix)
+        topic_name = safe_name(
+            context().prefix(), name, MAX_TOPIC_NAME_LENGTH, suffix=suffix, pulumi_suffix_length=0
+        )
 
         topic = sns.Topic(
             topic_name,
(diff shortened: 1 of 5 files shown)


stelvio/aws/topic.py
```diff
@@ -288,7 +288,9 @@ def topic_name(self) -> Output[str]:
     def _create_resources(self) -> TopicResources:
         suffix = FIFO_SUFFIX if self._fifo else ""
         name = self.name.removesuffix(suffix)
-        topic_name = safe_name(context().prefix(), name, MAX_TOPIC_NAME_LENGTH, suffix=suffix)
+        topic_name = safe_name(
+            context().prefix(), name, MAX_TOPIC_NAME_LENGTH, suffix=suffix, pulumi_suffix_length=0
+        )
```
sebst, line 293: would that change also be applicable to `Queue` (and potentially others)?

sebst (approved):

stelvio/aws/topic.py
michal-stlv: yes, same in Queue and the Cognito pools. Anywhere we pass an explicit name= the 8-char suffix reservation is pointless: harmless at SNS's 256 limit, could actually bite at SQS's 80. Will sweep those in a follow-up, keeping this PR small.

PR #262, opened 2026-08-24
fix: resolve default AWS region via the standard chain when not configured

michal-stlv:
With no `region` in `@app.config`, components that build region-based strings got a literal `None` — a hard deploy failure for `IdentityPool` (invalid provider name) and broken region values in AppSync DynamoDB data sources and CloudFront API origins. The region now resolves once via the standard AWS chain (`AWS_REGION` checked explicitly since boto3 ignores it, then `AWS_DEFAULT_REGION`/profile config); every provider is created with the resolved region, sites read their component's provider region via `aws_region_of()`, and an unresolvable region fails fast with a clear message. Adds unit coverage plus an integration test for the previously untested no-region path; verified live end-to-end (upgrade diff from an old-code stack shows zero updates/replacements).

Rebased on #264/#266. `WebsocketApi.url`/`management_url` read `context().aws.region` directly (rendering `execute-api.None` with no region configured) and now go through `aws_region_of()`, with no-region tests for both. `AcmValidatedDomain` picks its cross-region provider via `ProviderStore.aws_for_region()` alone; the old compare against the raw config region was redundant, since `aws_for_region` already returns the main provider on a match. The http_api test gate from this branch was dropped for the one #266 landed on main, which also covers owned `domain_name=` domains.

### DIFF
--- a/stelvio/aws/__init__.py
+++ b/stelvio/aws/__init__.py
@@ -1,1 +1,19 @@
 """AWS components for Stelvio."""
+
+from stelvio.provider import ProviderStore
+
+__all__ = ["default_region"]
+
+
+def default_region() -> str:
+    """Get the app's default AWS region.
+
+    The region set in `AwsConfig(region=...)`, or — when not set — the one resolved
+    from the standard AWS chain (`AWS_REGION`, profile config). This is the region
+    all Stelvio resources deploy to unless a component says otherwise.
+
+    Raises:
+        StelvioValidationError: If no region is configured anywhere.
+        RuntimeError: If called outside a Stelvio deployment operation.
+    """
+    return ProviderStore.region()
--- a/stelvio/aws/__init__.py
+++ b/stelvio/aws/__init__.py
@@ -9,8 +9,8 @@ def default_region() -> str:
     """Get the app's default AWS region.
 
     The region set in `AwsConfig(region=...)`, or — when not set — the one resolved
-    from the standard AWS chain (`AWS_REGION`, profile config). This is the region
-    all Stelvio resources deploy to unless a component says otherwise.
+    from the standard AWS chain (`AWS_REGION`, `AWS_DEFAULT_REGION`, profile config).
+    This is the region all Stelvio resources deploy to unless a component says otherwise.
 
     Raises:
         StelvioValidationError: If no region is configured anywhere.
--- a/stelvio/provider.py
+++ b/stelvio/provider.py
@@ -7,11 +7,20 @@ instead of relying on the implicit default provider.
 
 from __future__ import annotations
 
+import os
 from typing import TYPE_CHECKING, ClassVar
 
+import boto3
 import pulumi_aws
 
+from stelvio.exceptions import StelvioValidationError
+
 if TYPE_CHECKING:
+    from typing import Any
+
+    import pulumi
+
+    from stelvio.component import Component
     from stelvio.context import AppContext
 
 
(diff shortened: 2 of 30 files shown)


stelvio/provider.py
```diff
@@ -115,7 +173,12 @@ def _create_aws_provider(
...
-            region=region_override or ctx.aws.region,
+            region=region_override or cls.region(),
             profile=ctx.aws.profile,
             default_tags=pulumi_aws.ProviderDefaultTagsArgs(tags=all_tags),
         )
+
+
+def aws_region_of(component: Component[Any, Any]) -> str:
+    """Plain-str region the component's AWS provider deploys to."""
+    return ProviderStore.region_of(component._provider)  # noqa: SLF001
```
sebst, line 184: should we have a public property `provider`?

stelvio/aws/__init__.py
```diff
@@ -1 +1,19 @@
...
+from stelvio.provider import ProviderStore
+
+__all__ = ["default_region"]
+
+
+def default_region() -> str:
+    """Get the app's default AWS region.
+
+    The region set in `AwsConfig(region=...)`, or — when not set — the one resolved
+    from the standard AWS chain (`AWS_REGION`, profile config). This is the region
```
sebst: `AWS_DEFAULT_REGION` should be mentioned here as per the docs

stelvio/aws/__init__.py
```diff
@@ -1 +1,19 @@
 """AWS components for Stelvio."""
+
+from stelvio.provider import ProviderStore
+
+__all__ = ["default_region"]
```
sebst, line 5: nit: introduces different behaviour for `from stelvio.aws import *`.

stelvio/provider.py
```diff
@@ -31,6 +40,40 @@ class ProviderStore:
...
+        # (AWS_DEFAULT_REGION, profile config files).
+        region = (
+            os.environ.get("AWS_REGION") or boto3.Session(profile_name=ctx.aws.profile).region_name
+        )
+        if not region:
+            profile_hint = f" (profile: {ctx.aws.profile!r})" if ctx.aws.profile else ""
+            raise StelvioValidationError(
+                f"No AWS region configured{profile_hint}. Set one via "
+                "AwsConfig(region=...) in @app.config, the AWS_REGION environment "
+                "variable, or your AWS profile."
```
sebst: should also mention `AWS_DEFAULT_REGION`

sebst (changes requested):
all nits, but one: error message should mention AWS_DEFAULT_REGION as in docs

stelvio/provider.py
michal-stlv: done, 6c6f4b9

stelvio/aws/__init__.py
michal-stlv: done, same commit

stelvio/provider.py
michal-stlv: keeping it private for now. no user needs it yet and Component is cloud-agnostic, the provider is plumbing. if a real use shows up (raw pulumi resources under a component) we expose it then.

stelvio/aws/__init__.py
michal-stlv: intentional. `__all__` is how we control what `from stelvio.aws import *` gives you: just `default_region`, not the `ProviderStore` import or whatever submodules happen to be loaded. nobody star-imports stelvio.aws today so nothing changes in practice, but this way we decide what's exported, not python.

sebst (approved):

PR #265, opened 2026-08-26
fix: remove deprecated provider properties and validate DynamoTable fields

michal-stlv:
Three fixes. Together they remove every deprecation warning `stlv deploy` prints today.

**DynamoTable global indexes** move off the deprecated `hash_key`/`range_key` to `key_schemas`. Table-level keys and local-index `range_key` are not deprecated, so they stay as they are.

**Bucket versioning** moves off the deprecated inline `versioning` property to a separate `BucketVersioning` resource, created only when `versioning=True`. `"Disabled"` is valid only for a bucket that was never versioned, and config alone cannot tell us that. So nothing is created when versioning is off, and deleting the resource makes the provider suspend versioning, which is the right outcome for `True` to `False`.

**DynamoTable `fields`** now rejects entries that no key uses. It raises when the table is defined instead of failing mid-deploy with an AWS `ValidationException`.

### Verified against AWS

Deployed two tables (one GSI with a sort key, one without) and three buckets using the released code, then ran `stlv diff` on this branch: zero replacements, zero updates, one resource created, no warnings. The `key_schemas` change produces no diff at all.

Also confirmed by hand: toggling versioning off leaves the bucket `Suspended` rather than `Disabled`, and an unversioned bucket has no versioning configuration at all.

Integration tests now cover all three versioning states plus the suspend transition. The 6 relevant integration tests pass against AWS.

### Docs

This also fixes the docs. The new validation broke four published examples that declared non-key fields, so those are corrected. Fixing them turned up six more that were already broken. `AttributeType` was renamed to `FieldType` in #34 and the docs were never updated, so the quickstart has raised `ImportError` for about a year. Found by scripting a check over every Python example in `docs/`.


stelvio/aws/s3/s3.py
```diff
@@ -432,13 +432,24 @@ def _create_resources(self) -> BucketResources:
...
                 inject_tags=True,
             ),
             opts=self._resource_opts(),
         )
 
+        # "Disabled" is only valid for a bucket that was never versioned, and config
+        # alone can't tell us that. Deleting this resource suspends versioning instead.
+        # Deliberately not in BucketResources and not customizable: nothing on it is worth
+        # reading, so test_customization_sync.py can't see it either.
+        if self.versioning:
```
sebst, line 455: could still use the deprecated versioning via customize: `customize={"bucket": {"versioning": ...}}`. Maybe add a note in the docs

stelvio/aws/s3/s3.py
```diff
@@ -432,13 +432,24 @@ def _create_resources(self) -> BucketResources:
...
-                    "versioning": {"enabled": self.versioning},
                 },
                 inject_tags=True,
             ),
             opts=self._resource_opts(),
         )
 
+        # "Disabled" is only valid for a bucket that was never versioned, and config
+        # alone can't tell us that. Deleting this resource suspends versioning instead.
+        # Deliberately not in BucketResources and not customizable: nothing on it is worth
```
sebst, line 453: maybe obscure use case, but there is `versioning_configuration.mfa_delete` and `mfa`

stelvio/aws/s3/s3.py
michal-stlv: right, the argument is still on the Bucket resource so customize can reach it. but then pulumi prints its own deprecation warning, so the user gets told. customize is for people who know what they're doing, I'd leave it and skip the docs note.

stelvio/aws/s3/s3.py
michal-stlv: yeah I looked at it before and decided to skip. `mfa` needs the current TOTP code, which expires in 30s, so every deploy would need a fresh one typed in, and only root can enable mfa delete. not a flow for `stlv deploy`. if someone needs it: `versioning=False` and their own `pulumi_aws.s3.BucketVersioning` on `bucket.resources.bucket.id`.

docs/docs/changelog.md
```diff
@@ -6,6 +6,9 @@
...
 - **Default AWS region resolution.** With no `region` in `@app.config`, Stelvio now resolves the region once via the standard AWS chain (`AWS_REGION`, `AWS_DEFAULT_REGION`, profile config) and uses it everywhere. Previously a missing region put a literal `None` into region-based values, breaking `IdentityPool` deploys and AppSync DynamoDB data sources — and setting only `AWS_REGION` (which boto3 ignores) didn't work. When no region can be resolved at all, Stelvio fails fast with a clear error.
+- **`DynamoTable` global index deprecation warnings.** Global secondary indexes no longer print `hashKey is deprecated` / `rangeKey is deprecated` on every deploy and diff. They now use the provider's `key_schema` form. Existing tables see no diff, because the key schema is identical.
+- **`Bucket` versioning deprecation warning.** Buckets no longer print `versioning is deprecated`. Versioning is now a separate `BucketVersioning` resource, created only when `versioning=True`. Existing buckets see no diff. Setting `versioning=False` on a versioned bucket suspends versioning, as S3 can't make a bucket unversioned again.
```
sebst: existing unversioned buckets see no diff; existing versioned buckets gain a BucketVersioning resource (no replacement, no bucket update).

docs/docs/changelog.md
```diff
@@ -6,6 +6,9 @@
...
 - **Default AWS region resolution.** With no `region` in `@app.config`, Stelvio now resolves the region once via the standard AWS chain (`AWS_REGION`, `AWS_DEFAULT_REGION`, profile config) and uses it everywhere. Previously a missing region put a literal `None` into region-based values, breaking `IdentityPool` deploys and AppSync DynamoDB data sources — and setting only `AWS_REGION` (which boto3 ignores) didn't work. When no region can be resolved at all, Stelvio fails fast with a clear error.
+- **`DynamoTable` global index deprecation warnings.** Global secondary indexes no longer print `hashKey is deprecated` / `rangeKey is deprecated` on every deploy and diff. They now use the provider's `key_schema` form. Existing tables see no diff, because the key schema is identical.
```
sebst: `key_schemas`

stelvio/aws/s3/s3.py
sebst: nit: does not mention suspend-on-disable as in docs

stelvio/aws/dynamo_db.py
```diff
@@ -200,6 +204,35 @@ def __post_init__(self) -> None:
                     f"sort_key '{global_index.sort_key}' not in fields list"
                 )
 
+        self._validate_fields_are_keys()
+
+    def _validate_fields_are_keys(self) -> None:
```
sebst, line 209: should have a test that checks that this table is invalid:

```
DynamoTable(
    "users",
    fields={"id": "S", "email": "S"},
    partition_key="id",
    global_indexes={
        "by-id": GlobalIndex(partition_key="id", projections=["email"]),
    },
)
```

`email` is on the GSI as a projected attribute, not as a key. AWS would reject it. The new validator is supposed to catch that.

tests/aws/test_customization.py
sebst, line 381: nit: since versioning is now not an attribute but a `BucketVersioning` resource, this tests does not actually test merging

stelvio/aws/dynamo_db.py
```diff
@@ -200,6 +204,35 @@ def __post_init__(self) -> None:
...
+
+        for index in self.global_indexes.values():
+            global_index = index if isinstance(index, GlobalIndex) else GlobalIndex(**index)
+            key_fields.add(global_index.partition_key)
+            if global_index.sort_key:
+                key_fields.add(global_index.sort_key)
+
+        unused_fields = [f for f in self.fields if f not in key_fields]
+        if unused_fields:
+            raise ValueError(
```
sebst, line 230: nit: maybe add table name to validation message?

sebst (changes requested):

docs/docs/changelog.md
michal-stlv: right, our own diff showed the 1 create on the versioned bucket. took your wording.

docs/docs/changelog.md
michal-stlv: fixed. I copied `key_schema` from the deprecation message, which names the terraform block. the pulumi input is `key_schemas`.

stelvio/aws/s3/s3.py
michal-stlv: added, same sentence as in the docs.

stelvio/aws/dynamo_db.py
michal-stlv: the validator already rejects it, it only collects keys, so `email` lands in the unused list. added your table as a `projected-field-is-not-a-key` case in `test_dynamo_table_rejects_fields_not_used_as_keys`.

tests/aws/test_customization.py
michal-stlv: yeah, once `versioning` left the bucket resource those three tests only checked the bucket name. moved them to `Function(memory=256)` + customize `timeout`, so a param and a customize value have to land on the same resource again. same for the empty/None customize tests below it.

stelvio/aws/dynamo_db.py
michal-stlv: did it one level up: `DynamoTable.__init__` wraps the config error as `DynamoTable 'users': ...`, so every config message gets the name, not just this one. a config the user builds directly stays as is, it has no name yet.

sebst (approved):

PR #267, opened 2026-09-10
tests: unflake api gateway tests

sebst:


michal-stlv (approved):
okay, all good, approving. ran the api gateway unit tests in a loop locally, all green.
two notes, no changes needed:
1. the websocket stage depends_on is not needed by any test. the ws tests already assert after the inner deploy() drain. it only orders the stage after the permissions on real aws, which is harmless.
2. the http side works because the list lands in HttpApiResources.permissions and when_http_api_ready waits on it. we plan to move the http tests to the same inner deploy() pattern as the ws tests, then that gate goes away. (also as agreed permisson won't be in Resources class)

PR #271, opened 2026-09-10
fix: friendly message for missing or rejected AWS credentials

michal-stlv:
Any `stlv` command run with unconfigured or rejected AWS credentials ended in a 60-line botocore traceback with the `Loading app...` spinner left on screen. It now stops with a three-line red message and exit code 2 (same text under `--json`): botocore's own error as the headline, which profile Stelvio used and where it came from (`AwsConfig(profile=...)`, `AWS_DEFAULT_PROFILE`, `AWS_PROFILE`, or the default when none is set), and a fix hint. `STLV_DEBUG=1` re-raises the raw exception. The translation is a context manager in `stelvio/aws/home.py` around the three places every command hits first (region resolution, `AwsHome` construction, the first `read_param`); it only catches credential-chain errors and SSM's four auth error codes, so anything else (throttling, a deleted state bucket, an unreachable endpoint) keeps its traceback. The spinner is a context manager too now, so a failing `CommandRun.__enter__` stops it.

Tests: unit coverage for each message shape (no creds, partial creds, unknown or empty profile via either variable, SSO profile without a cached token, each SSM code, plus the pass-through cases), a spinner test, and one integration test that hits real SSM with a bogus key to pin `UnrecognizedClientException`. Also refreshes the troubleshooting page and syncs the `uv.lock` version line left over from the 0.10 release.


docs/docs/intro/troubleshooting.md
```diff
@@ -71,12 +71,27 @@ Only use this if you're certain no other deployment is actually running.
 
 ### AWS Credential Issues
 
-**Problem:** "Unable to locate credentials" or "Invalid security token"
+**Problem:** A `stlv` command stops with a short red message like this:
```
sebst, line 74: this would be a good place to introduce `stelvio` as the default command (not renaming everything, just the cli call) so that it matches the package name

michal-stlv: Well, I would say it needs to be everywhere or nowhere. If it's just in this one place and it's original everywhere else, it's kind of strange.

docs/docs/changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.11.0b7 (2026-MM-DD)
+
+### Bug Fixes
+
+- **Friendly AWS credential errors.** Missing credentials, an unknown profile, an expired SSO session, or a rejected key now stop `stlv` with a short message and a fix hint instead of a traceback. → [Troubleshooting](intro/troubleshooting.md#aws-credential-issues)
```
sebst: `→ ` plus link should be new paragraph

stelvio/aws/home.py
```diff
@@ -1,27 +1,128 @@
...
+        "UnrecognizedClientException",  # access key id does not exist
+        "InvalidSignatureException",  # secret key is wrong
+    }
+)
+# botocore's order: AWS_DEFAULT_PROFILE wins over AWS_PROFILE.
+_PROFILE_VARS = ("AWS_DEFAULT_PROFILE", "AWS_PROFILE")
+
+
+@contextmanager
+def translate_aws_errors(profile: str | None, region: str | None) -> Iterator[None]:
```
sebst: "translate" here is a bit misleading imho. maybe `convert_aws_errors`?

docs/docs/changelog.md
michal-stlv: fixed

stelvio/aws/home.py
michal-stlv: i like translate but if not clear i understand. renamed now.

tests/test_provider.py
```diff
@@ -46,6 +46,29 @@ def test_default_region_unresolvable_raises_friendly_error(no_region_context, mo
         default_region()
 
 
+def test_default_region_with_unknown_profile_raises_friendly_error(no_region_context, monkeypatch):
```
sebst, line 49: no profile, but default region and access_key in env would still work?

from 0.10 changelog: With no `region` in `@app.config`, Stelvio now resolves the region once via the standard AWS chain (`AWS_REGION`, `AWS_DEFAULT_REGION`, profile config) and uses it everywhere.

docs/docs/intro/troubleshooting.md
sebst: I'm suggesting to do it everywhere in this PR optionally, but fine with me to leave for follow up

michal-stlv: We can do it everywhere, but I would definitely not do it in this PR.

tests/test_provider.py
michal-stlv: yes it still works. this test is for profile set in AwsConfig that doesn't exist, that one must fail.

no profile = boto3 normal chain, region from env (test above), keys from env. we only wrap errors, if no error nothing changes. i checked it, no profile + AWS_DEFAULT_REGION + keys in env only works fine. integration test in tests/integration/test_home.py does the same (no profile, keys in env).

sebst (approved):

sebst (commented):
all good

PR #273, opened 2026-09-14
test: per-test event loop, mocks reset after each test, hermetic aws

michal-stlv:
Three unit-suite fixes in `tests/conftest.py`, from tests that passed in serial CI and failed alone or under xdist.

- Own asyncio loop per test, cancelled and closed after it. Constructing a Component registers a `pulumi.ComponentResource`, which queues async work on the current loop; a validation-only test never pumps it, so the work ran inside the next test that did, against its mocks. 712 test runs ended that way.
- `pulumi_mocks` calls `reset_options()` on teardown, and the conftest calls it once at import so a fresh process starts in the same state. A test that drives resources without the fixture now fails every time with Pulumi's `create_urn` TypeError instead of riding the previous test's mocks.
- `hermetic_aws` runs for every test outside `tests/integration/`, so no unit test reads the machine's `~/.aws` (why the #262 CI failure was invisible locally).

With one fixture doing the reset, the other three `set_mocks` callers fold into it: `EmailTestMocks` (SES DKIM tokens and ConfigurationSet arn move to `OUTPUT_TEMPLATES`), `_MinimalMocks` + `clear_registry` in the component and link tests, the duplicate `pulumi_mocks` in `tests/aws/conftest.py`. Module-level loops in the bridge tests go too.

Tests only, no changelog. Green on 3.12 and 3.14; the 19 formerly order-dependent tests pass alone. Suite time is unchanged against main: the guard first requested a `tmp_path` per test, which cost 2 seconds, so it now uses one shared empty HOME for the session.


tests/conftest.py
```diff
@@ -105,10 +125,17 @@ def hermetic_aws(monkeypatch, tmp_path):
         "AWS_SESSION_TOKEN",
     ):
```
sebst, line 128: could add `AWS_REGION` and `AWS_DEFAULT_REGION`

sebst (dismissed):

tests/conftest.py
michal-stlv: good point. applied. you need to reapprove both prs now though

sebst (approved):

PR #277, opened 2026-09-14
feat: lambda in vpc

michal-stlv:
## Why

Functions that talk to a database inside a VPC (DocumentDB is next) have to run
inside that VPC. Stelvio had `Vpc` but no way to put a `Function` in it.

### `Function(vpc=...)`

- `Function("worker", handler=..., vpc=vpc)` puts the function in the Vpc's
  private subnets (one per AZ) and attaches `AWSLambdaVPCAccessExecutionRole`
  next to the basic execution policy.
- `VpcAttachment(vpc=, subnets="private" | "isolated", security_groups=[...])`,
  or the same as a dict, for control. Public subnets are rejected: Lambda network
  interfaces never get a public IP, so a public subnet gives no route out.
  `security_groups` takes up to 5 existing group ids and replaces the app group;
  those users wire their own rules.
- `vpc` is a `FunctionConfig` field, so API route handlers and every other
  component that creates a Function get it too.
- Dev mode: the stub Lambda stays outside the VPC (it has to reach the websocket
  bridge). The app group and the VPC policy stay in place, so switching between
  `stlv dev` and `stlv deploy` is an in-place update with no network-interface
  wait.

### One app security group per Vpc

- `Vpc` creates `{vpc}-app-sg` the first time a function attaches: no inbound
  rules, one standalone egress rule for all traffic. Every attached function
  shares it. Datastore components will open their port to it (next PR).
- This follows AWS's own Lambda guidance: reuse subnet + security group
  combinations. One set of network interfaces per Vpc instead of per function,
  no pressure on the 250-interface quota, no per-function destroy waits.
- Rules are standalone resources, never inline: the provider strips AWS's
  default egress rule on create, and inline rules cannot coexist with standalone
  rule resources on one group. So the app group has no customize key on
  purpose; a different rule set means bringing your own group.
- Not in `VpcResources`; nothing outside the library needs the id yet.

### Default security group emptied

`Vpc` adopts the VPC's default security group with no rules
(`ec2.DefaultSecurityGroup`), so AWS's allow-all rules go away on the first
deploy. Nothing Stelvio creates uses that group, and leaving it open is a
CIS 5.4 finding.

### Small things

- CLI labels for `Security Group`, `Security Group Egress Rule`,
  `Default Security Group`.
- `test_component_tagging_contract` now asserts after settlement instead of
  inside the trigger's `apply()`, which raced resources the trigger did not
  depend on (it flaked once in CI on the egress rule). The egress rule and the
  default group are in the contract now.

## Tests

- Unit: `tests/aws/function/test_function_vpc.py` (attachment forms, tiers,
  own groups, 5-group cap, dev-mode stub, HttpApi route handler, validation),
  `tests/aws/test_vpc.py` (app group, egress rule, default group with exact
  inputs, sealed counts).
- Integration: `test_function_in_vpc` deploys a NAT-less Vpc with a private and
  an isolated function plus a control outside the VPC. A probe handler proves
  the two inside have no internet egress and the control has, then reads back
  subnets, the shared group id and its rules. `test_vpc_default` checks the
  default group is empty, `test_vpc_tags` its tags. Full standard tier green.
- Manual on real AWS: deploy, `stlv dev`, redeploy, detach and attach, NAT on,
  destroy; a second run after the default-group change.

## Docs

`vpc.md` gains "Security Groups" and "Lambda Functions in VPC", `lambda.md`
"Running in a VPC", changelog entry under 0.11.0b7.

## Notes

- The first function deployed into a Vpc takes a few minutes (network
  interface creation); later functions reuse the interfaces. Destroying the Vpc
  waits for Lambda to release them.
- Not in this PR: datastore-side security groups and their ingress rule from
  the app group (comes with DocumentDB), a reachability test through NAT, VPC
  endpoints, dev-mode access to VPC resources.


stelvio/aws/vpc.py
```diff
@@ -183,6 +200,35 @@ def _create_resources(self) -> VpcResources:
...
         )
 
+    @cached_property
+    def _app_security_group(self) -> SecurityGroup:
+        # Shared security group every attached function wears; datastore components source
+        # their one ingress rule from it. Made on first read (the first attaching function),
+        # never from _create_resources: `self.resources` there would recurse.
+        # AWS security group names are capped at 255, unlike the other Vpc children which
+        # carry only a Name tag.
+        sg_name = safe_name(context().prefix(), self.name, 255, "-app-sg")
```
sebst, line 210: should this call `self._safe_name`, too?

stelvio/aws/vpc.py
```diff
@@ -183,6 +200,35 @@ def _create_resources(self) -> VpcResources:
...
+            description="Stelvio app tier: shared by functions attached to this VPC",
+            tags={"Name": sg_name} | self.tags,
+            opts=self._resource_opts(),
+        )
+        # No inline rules, ever: the provider drops AWS's default allow-all egress on
+        # create, and inline rules can't coexist with standalone rule resources on one
+        # group (the provider reverts one or the other). Datastores add standalone
+        # ingress rules later, so egress is standalone too.
+        SecurityGroupEgressRule(
+            context().prefix(f"{self.name}-app-sg-egress"),
```
sebst, line 223: safe_name?

sebst (approved):
two nits

stelvio/aws/vpc.py
michal-stlv: no, on purpose. security group is the only vpc child with a real aws name, pulumi adds its 8 char suffix and aws caps it at 255. `_safe_name` is 256 with no suffix, for the children that only have a Name tag.

stelvio/aws/vpc.py
michal-stlv: no, egress rule has no name in aws, it's only the pulumi urn. same as the routes below and acm/s3.

PR #278, opened 2026-09-16
fix: route-based api gateway child names, labelled cli child lines

michal-stlv:
## Why

Two things, one branch.

1. `stlv` printed same-type children with no way to tell them apart. A two-AZ
   Vpc with NAT showed six identical `Subnet` and `Route Table` lines; a
   RestApi showed a stack of bare `API Method` lines.
2. REST methods, integrations and resources, and HTTP/WebSocket routes, were
   named from a flattened path: braces dropped, segments joined with `-`.
   Distinct routes got the same Pulumi name and the deploy failed with a
   duplicate URN error:
   - `/user-profiles` and `/user/profiles`
   - `/users/{id}` and `/users/id`
   - `/users/` and `/users` under CORS (two OPTIONS sets for one resource)

### Same-type children say which one they are

- Under one component, children that share a type print a label:
  `Subnet (public-a)`, `IAM Policy Attachment (default)`,
  `Lambda Permission (functions-todos_get)`. The label is the child's logical
  name minus the component's name. A component that wants a shorter one
  registers a `@child_label("Type")` function (Vpc, RestApi, HttpApi,
  WebsocketApi do). Lone children stay bare.
- `stlv diff` and `stlv refresh` sort a component's children by type,
  sub-components first, API paths as a tree. Deploy and destroy keep arrival
  order.
- `aws:cloudwatch/logGroup:LogGroup` gets a display name. HTTP and WebSocket
  APIs show as `HTTP API` / `WebSocket API` instead of `API Gateway` (both are
  one Pulumi type; the protocol comes from the resource inputs, old side on
  destroy). API Gateway v2 authorizers get a label instead of the raw type.

### API Gateway children are named after their route

- `api-method-GET /users/{id}`, `api-resource-/users`, `http-route-$default`,
  `ws-route-$connect`. Fixes the collisions above. Old names stay as Pulumi
  aliases, so existing stacks migrate in place on the next `stlv deploy`,
  nothing is replaced. The aliases can be dropped in a later release once
  every stack has deployed on this one.
- REST paths are canonicalized at definition: `/users/` and `/users` are one
  route (one conflict check, one CORS set). A stack that already had a
  trailing-slash route gets one API redeployment, no resource change.
- With the route in the name, the label shows it:
  `API Method (GET /users/{id})`, `API Route ($connect)`.

Docs: changelog, contributing guide (`@child_label`), quickstart sample output.


tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -787,3 +787,23 @@ def test_http_api_access_log_retention_invalid_boundaries(days):
...
+    api.route("ANY", "$default", "functions/simple.handler")
+
+    def check(_):
+        assert set(old_names) == {
+            f"{TP}my-api-route-GET--users-{{id}}",
+            f"{TP}my-api-route-GET",
+            f"{TP}my-api-route-default",
+        }
+
+    when_http_api_ready(api, check)
```
sebst, line 809: maybe add a test like this 

```python
@pulumi.runtime.test
def test_rest_api_routes_that_used_to_share_a_name_are_distinct(pulumi_mocks):
    api = RestApi("test-api")
    api.route("GET", "/user-profiles", handler=Funcs.SIMPLE.handler)
    api.route("GET", "/user/profiles", handler=Funcs.SIMPLE.handler)
    api.route("GET", "/users/{id}", handler=Funcs.SIMPLE.handler)
    api.route("GET", "/users/id", handler=Funcs.SIMPLE.handler)
    def check(_):
        methods = {
            m.name
            for m in pulumi_mocks.created_methods()
            if m.inputs["httpMethod"] == "GET"
        }
        assert methods == {
            f"{TP}test-api-method-GET /user-profiles",
            f"{TP}test-api-method-GET /user/profiles",
            f"{TP}test-api-method-GET /users/{{id}}",
            f"{TP}test-api-method-GET /users/id",
        }
    when_api_ready(api, check)
```

michal-stlv: Added in aa7a8ed, right after the alias test. Methods only: those are the names that collided, and the mocks have no Resource-listing helper yet. Also merged main (#277 and #280) so the conflicts are gone.

sebst (approved):

PR #279, opened 2026-09-16
test: tighten weak asserts, rewrite asserts in integration helpers

michal-stlv:
## Why

Five asserts that could not fail, or would pass on the wrong value:

- `test_http_api.py` imported the component's private `_ACCESS_LOG_FORMAT` and
  compared it to itself.
- Two DynamoTable validation tables tested the config class, not the component,
  and shared two identical cases.
- The two DNS tests pinned the ApiDomain target with
  `endswith(".execute-api.{region}.amazonaws.com")`, which any regional API
  Gateway hostname satisfies.
- `test_auto_tags` checked the region as a substring of the ARN.

And a failing `assert` inside `tests/integration/assert_*.py` printed nothing,
because pytest rewrites asserts only in test files and conftest. That is why
nearly every helper assert carried a hand-written message.

## What

- The log format literal is inlined in the test.
- One DynamoTable validation table, driven through `DynamoTable(...)`, with the
  constructor's `DynamoTable 'test': ` prefix pinned. The one case only the
  deleted table had moved next to its siblings.
- New `assert_apigatewayv2_domain` helper: reads the domain back and asserts
  exactly one configuration, `REGIONAL`, `TLS_1_2`, at the exported target.
- Region asserted as the ARN's region field.
- `tests/conftest.py` registers `tests.integration` for assert rewriting, so
  helper asserts print both sides on failure. 93 messages that only restated
  the operands are gone; 34 that add something the expression lacks (the
  resource id passed in, the candidate list behind a `len()`) stay. Assert
  expressions are unchanged, checked by parsing both versions.

Unit suite green. `test_auto_tags` and both DNS tests run green against AWS.


tests/aws/api_gateway/http_api/test_http_api.py
```diff
@@ -197,7 +196,17 @@ def verify_http_api(mocks, case: HttpApiTestCase) -> None:
         {
             "name": case.stage_name,
             "accessLogSettings": {
-                "format": _ACCESS_LOG_FORMAT,
```
sebst, line 200: nit: could be named copy

tests/integration/test_provider.py
```diff
@@ -44,8 +44,8 @@ def infra():
 
     # Resources are in the expected region
     region = stelvio_env.aws_region
-    assert f":{region}:" in outputs["queue_tasks_arn"]
-    assert f":{region}:" in outputs["dynamotable_orders_arn"]
+    assert outputs["queue_tasks_arn"].split(":")[3] == region
+    assert outputs["dynamotable_orders_arn"].split(":")[3] == region
```
sebst, line 48: If changed here, consider changing in `tests/integration/test_dns_cross_region.py`, line 52, too

sebst (dismissed):

tests/aws/api_gateway/http_api/test_http_api.py
michal-stlv: Done in e3454b9: ACCESS_LOG_FORMAT constant at the top of the file.

tests/integration/test_provider.py
michal-stlv: Done in e3454b9, both region asserts in that file (line 53 queue ARN, line 67 cert ARN) now check the region slot.

sebst (approved):

PR #283, opened 2026-09-18
feat: AWS `DocumentDb` component **DRAFT**

sebst:
some notes following our discussion:

- support for username/password (without secrets manager) in a follow up (#282)
- auto bundler included (pem file injected through linking mechanism). deliberately chose to not vendor the pem file (to reduce maintenance burden)
- fc8b9a9234d461edf179728e24907b10ab2fa5e6 changed the mongodb connection string back to follow rotation pattern via secrets manager (in docs); re-added in c111b6b5c4bc4b76b5dae562169fab1022b17263 with the ability to disable rotation; later 30b4ecfbbdda3ca1021e029eec827c30cfb3bc7e removes connection_string entirely in favor of connection_uri
- adapted #284 naming guides in 0e9b084994af30a7af4e61cb027055b6eecc039a
- adapted #285, #286 in b5f86e2389afee978fc6936ced504057abd2be6b
- bumped pulumi version in 22e69cebdcfbb194b153e3754ddad76116649eee, then added support for native `rotation_enabled=False` in 4e58d3e2c0e85d315c25f67969ac56c06b50244f


docs/docs/components/aws/document-db.md
```diff
@@ -0,0 +1,372 @@
...
+billed per request, start with [`DynamoTable`](dynamo-db.md).
+
+By default, `DocumentDb` creates a private, TLS-required, encrypted cluster on
+engine `8.0` with an AWS-managed master password. The cluster sits in the
+Vpc's isolated subnets and accepts traffic from that Vpc's shared app security
+group.
+
+!!! warning "DocumentDB costs money"
+    Instances bill while they run, not per request. One `t4g.medium` is
+    **~$47/month** ($0.065/hour × ~730 hours in `us-east-1`) before storage and
```
michal-stlv, line 18: The hourly price looks off. AWS lists `db.t4g.medium` in `us-east-1` at $0.07566/hour. With 730 hours that is ~$55/month, not ~$47. Same on line 263. The "~$50" typical cluster goes up too.

docs/docs/components/aws/document-db.md
```diff
@@ -0,0 +1,372 @@
...
+You can also pass a dict via `config=`. Don't mix kwargs and `config=` in the
+same call.
+
+Available configuration options:
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| `vpc` | (required) | Existing `Vpc`. The cluster uses its isolated subnets. |
+| `instances` | `1` | Number of cluster instances (1 to 16). Extra instances are replicas. |
+| `instance_class` | `None` | Instance size, with or without the `db.` prefix (`"t4g.medium"` or `"db.t4g.medium"`). `None` uses a global `instance` customize default if set, otherwise `t4g.medium`. |
```
michal-stlv, line 81: Can we just say "`None` uses `t4g.medium`"? The global customize rule works the same for every default. Mentioning it only here is confusing.

docs/docs/components/aws/document-db.md
```diff
@@ -0,0 +1,372 @@
...
+Default is `"8.0"`. Pass `engine="5.0"` to opt in to 5.0.
+
+!!! warning "Changing engine upgrades the live cluster"
+    Changing `engine="5.0"` to `engine="8.0"` is an in-place major version
+    upgrade with downtime. Stelvio blocks that unless you opt in with
+    `customize={"cluster": {"allow_major_version_upgrade": True}}`. You must
+    still meet the
+    [AWS upgrade prerequisites](https://docs.aws.amazon.com/documentdb/latest/devguide/docdb-mvu.html).
+    You cannot downgrade by changing `engine` back. AWS will not major-upgrade a
+    `t4g.medium` writer. Resize first, in a separate deploy.
```
michal-stlv, line 112: This is not enough on its own. Stelvio does not set `apply_immediately`, so the resize waits for the next maintenance window. If the user changes `engine` right after, the upgrade still hits the `t4g.medium` writer.

Either the resize uses `apply_immediately`, or the docs need real steps for the user. The first draft had the `apply_immediately` sentence.

docs/docs/components/aws/document-db.md
```diff
@@ -0,0 +1,372 @@
...
+can reach the cluster. See
+[Lambda Functions in VPC](vpc.md#lambda-functions-in-vpc).
+
+A Function that links a `DocumentDb` must set `vpc=` to the **same** `Vpc` as
+the cluster. Missing `vpc=` or a different Vpc raises `ValueError` when the
+Function is created. Linking injects connection properties and IAM; it is not
+networking.
+
+The `connection_uri` property does not call Secrets Manager at runtime, so
+isolated-subnet Functions can talk to the cluster without NAT.
```
michal-stlv, line 128: This is not right. `connection_uri` has no password. The function still needs the password from Secrets Manager, so it still needs NAT. Please remove "without NAT".

docs/docs/components/aws/document-db.md
```diff
@@ -0,0 +1,372 @@
...
+`connection_uri` stays valid when the password rotates. If you fetch
+`secret_arn` at runtime, private subnets need NAT or a Secrets Manager VPC
+endpoint. Stelvio does not create the endpoint. `nat="managed"` only routes
+private subnets through NAT; isolated subnets stay isolated.
+
+Custom `security_groups` on `VpcAttachment` are kept. When the Function also
```
michal-stlv, line 134: From the code (`_vpc_config` in `function.py`): when a function has its own `security_groups` and links a `DocumentDb`, Stelvio adds the app group. Please remove that. Links should not touch security groups.

A user who passes their own groups wants this function outside the shared app group. This puts it back in. Then it can reach every datastore that trusts the app group. Adding a link also changes the function's security groups, and AWS recreates the network interfaces for that.

So this paragraph should say the opposite: with your own `security_groups`, you add the rule yourself. Can we show a short example here? Something like:

```python
from pulumi_aws.vpc import SecurityGroupIngressRule

SecurityGroupIngressRule(
    "todos-from-my-sg",
    security_group_id=db.resources.security_group.id,
    referenced_security_group_id=my_sg.id,
    ip_protocol="tcp",
    from_port=27017,
    to_port=27017,
)
```

Right now `db.resources.security_group` is only mentioned in the customize warning at the end.

docs/docs/components/aws/document-db.md
```diff
@@ -0,0 +1,372 @@
...
+
+!!! info "Dev mode cannot reach the cluster yet"
+    `stlv dev` runs your handlers on your machine, outside the VPC, so the
+    cluster is not available in dev mode yet. Access is coming soon.
+
+## Linking
+
+Put the Function in the same Vpc and link it. Linking packages Amazon's
+[global RDS CA bundle](https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem)
+into the Function as `stlv_docdb_ca.pem` and injects that path as `ca_file`.
```
michal-stlv, line 148: Two things from the code behind this.

Where the file comes from (`_download_document_db_ca` in `document_db.py`): it is downloaded on deploy and cached for 24 hours. Why not ship it inside the `stelvio` package? Then there is no download code, no cache and no network call on `stlv deploy` or `stlv diff`. Now a deploy fails when the download fails and the cache is older than 24 hours. Root CAs change rarely, and a Stelvio release can update the file.

How it gets into the zip (`_document_db_ca_assets` in `function.py`): `Function` looks for `DocumentDb` links itself. I would prefer this to be generic. RDS or Aurora would need their own hook there later. Can the link bring the files instead? For example a `files` field on `LinkConfig`, which `Function` passes to `extra_assets`.

docs/docs/components/aws/document-db.md
```diff
@@ -0,0 +1,372 @@
...
+import json
+
+import boto3
+from pymongo import MongoClient
+from stlv_resources import Resources
+
+secrets = boto3.client("secretsmanager")
+
+
+def handler(event, context):
```
michal-stlv, line 231: The example fetches the secret and creates a new `MongoClient` on every call, and it never closes the client. What we agreed on the call:

- fetch the password and create the client at module level, so warm calls reuse them
- on an auth failure, fetch the new password, close the old client and reconnect once
- don't replay a write automatically, we don't know if it went through

The part about rotation and auth failures from the first draft is also gone. Can we bring it back?

docs/docs/components/aws/document-db.md
```diff
@@ -0,0 +1,372 @@
...
+    collection.replace_one({"_id": "hello"}, {"_id": "hello", "ok": True}, upsert=True)
+    return {"item": collection.find_one({"_id": "hello"})}
+```
+
+`host`, `port`, `ca_file`, and `replica_set` are still injected if you need the
+pieces.
+
+!!! info "DocumentDB is not full MongoDB"
+    TLS is required. The URI already sets `replicaSet=rs0` and `retryWrites=false`.
+    DocumentDB does not support retryable writes, and clients that omit the
```
michal-stlv, line 250: Can we drop "clients that omit the replica set name often fail to discover the cluster"? The URI already sets it, and pymongo 4 finds `rs0` by itself.

docs/docs/components/aws/document-db.md
```diff
@@ -0,0 +1,372 @@
...
+for your region.
+
+A default cluster bills:
+
+- Instance: one `t4g.medium` at **$0.065/hour**, about **~$47/month**
+- Storage: **$0.10/GB-month**
+- I/O: **$0.20 per million requests**
+- Secrets Manager storage for the managed password: about **~$0.40/month**
+
+NAT (**~$37 to ~$73/month** with `nat="managed"`) is needed for the Function to
```
michal-stlv, line 268: NAT is a VPC cost, not a DocumentDB cost. Can we drop the NAT prices here? One short line is enough: the function needs NAT to read the password, see [VPC cost](vpc.md#cost). Same short line in the cost box at the top.

sebst: done

docs/docs/components/aws/document-db.md
sebst: removed

docs/docs/components/aws/document-db.md
sebst: fixed

docs/docs/components/aws/document-db.md
sebst: done

docs/docs/components/aws/document-db.md
sebst: done

docs/docs/components/aws/document-db.md
sebst: fixed

docs/docs/components/aws/document-db.md
sebst: changed

docs/docs/components/aws/document-db.md
sebst: changed. original intention was: auto bundler included (pem file injected through linking mechanism). deliberately chose to not vendor the pem file (to reduce maintenance burden). validity is long enough but might get revoked, tho

docs/docs/components/aws/document-db.md
sebst: back in

docs/docs/components/aws/document-db.md
michal-stlv: sorry, claude messed up my notes here, this is not what i meant (see chat)

1. download - please put it back as you had it, with the cache. your reasoning was right. remove the pem from the package

2. files field - it's one option and if you think it's the best i can live with it. but i'd rather hack it into properties. stelvio puts there some special property e.g. STLV_INTERNAL_FILES, dict where key is file name in the lambda zip and value is path to the file. we need to define what the relative paths are relative to so it's reusable. dict so it can carry more files later, one file is enough now. Function reads it, packages the files and doesn't pass it as env var

sorry for the extra round

PR #284, opened 2026-09-18
feat: one naming rule, replacement-safe names for every resource

michal-stlv:
## Why

Stelvio had two naming patterns side by side: some resources got an explicit deterministic name (`Bucket`, `Queue`, `Topic`, `UserPool`, ...), others were left to Pulumi's autonaming (`Function`, `DynamoTable`, IAM). The split was accidental and it cost us:

- Deterministic names make replacements unsafe. When a property AWS can't change in place forces a new resource, Pulumi creates the new one before deleting the old, and with a fixed name the two collide. A random suffix makes every replacement create-before-delete for free, and a replacement always gets a fresh suffix.
- Names in shared namespaces (S3 is global) relied on nobody else using the same app and env prefix.
- Every explicit name still reserved 8 chars for a Pulumi suffix that never came, so long names truncated 8 chars early (found in #245).
- The prefix and the length guard were opt-in, and got skipped twice (#122, #230). Two of the six explicit-name sites had no length guard at all.

This PR settles on one rule: let Pulumi name the resource whenever the provider allows it, and route every AWS-facing name through one helper that adds the prefix and the guard. Same shape SST uses, without their transform hook, because Stelvio's logical names already carry the app-env prefix.

## What changes

New queues, topics, buckets, user pools, user pool clients and static-website CloudFront functions get Pulumi-generated names, `<app>-<env>-<name>-<random>`. Existing deployments keep their names: Pulumi reuses the stored name when the input goes away. Verified on real AWS twice: `stlv diff` on an environment deployed from main showed only the two FIFO replacements below, and a second run with a long-named, non-empty bucket showed no changes.

Nothing changes in `stlv_app.py`. Names come from env vars and links as before, so no user code reads them.

## Replaced on the next deploy

- FIFO topics, and any `Queue` or `Topic` named `*.fifo`. The logical name loses `.fifo` (the provider appends it after the random suffix, and a dot mid-name is invalid), which changes the URN. Subscriptions are recreated.
- Queues and topics whose prefix plus name passes 72 chars. The logical name now truncates, because Pulumi rejects an overflow at preview.
- The Email configuration set, now prefixed with app and env. Two apps in one account used to collide. Event destinations are recreated with it.

## Deterministic names, no random suffix

`IdentityPool`, `Layer` and the SES configuration set: the provider requires the name input. Identity pool names need not be unique, layer versions group under one layer name, and the config set's only replace trigger is its own name.

## Length guards

Every logical name now carries its real limit: SQS 80 (FIFO reserves 5 for `.fifo`), SNS 80 (pulumi-aws caps autonames at 80 although AWS allows 256), S3 63, CloudFront Function 64, Cognito 128. New: `Layer` at 80, because `CreateFunction.Layers` caps each version ARN at 140 chars, so a longer layer publishes but can never be attached. `Bucket` aliases its old unguarded logical name, so a long-named bucket deployed before this change keeps its URN and its data.

## Internals

`resource_name(base, *, limit, suffix="", pulumi_suffix_length=8)` in `stelvio.component` wraps `safe_name` with the app-env prefix. This PR uses it at the nine sites whose AWS name changes. The remaining `safe_name` sites (logical names, IAM, mappings, subscriptions) move to it in a follow-up PR, after which `safe_name` folds into it.

Tests: per site, no physical-name input plus truncation at each limit; queue and topic as full-input compares. `.fifo` on the real name is asserted by the integration helpers. Docs: new concept page `docs/docs/concepts/naming.md`; `contributing/components.md` §Naming rewritten around the three recipes. Read that section first.


docs/docs/changelog.md
```diff
@@ -8,14 +8,26 @@
...
+    → [Resource Naming](concepts/naming.md)
+
 ### CLI
 
 - **Same-type child resources show which one they are.** `Subnet (public-a)`, `API Method (GET /users/{id}/orders)`, `IAM Policy Attachment (default)` instead of six identical `Subnet` lines.
 - **Diff output is sorted.** `stlv diff` and `stlv refresh` group children by type, sub-components first, API paths as a tree. `stlv deploy` keeps arrival order.
 
+### Breaking Changes
+
+- **Replaced on the next deploy:** FIFO topics, any `Queue` or `Topic` named `*.fifo`, queues and topics whose prefix plus name passes 72 chars, and the Email configuration set (event destinations recreated with it).
```
sebst, line 25: nit: it's 67 for fifo queues

docs/docs/contributing/components.md
```diff
@@ -115,31 +115,41 @@ on taggable resources only, and keep it at the callsite where you can see it.
...
+- `limit`: where the string lands. The AWS limit for the resource type (63 for buckets,
+  128 for user pools), or the provider's own cap when that is lower: pulumi-aws cuts SQS
+  and SNS autonames at 80 even though SNS allows 256. For tag-only names, 256.
+- `pulumi_suffix_length`: keep the default 8 when Pulumi appends its suffix (recipe 1), so
+  the final AWS name still fits. 0 for recipes 2 and 3.
+- `suffix`: a tail that must survive truncation; it is re-appended after the hash. Only
+  needed for names Stelvio sets itself. For autonamed FIFO queues and topics the provider
+  appends `.fifo` after its random suffix, so the component strips `.fifo` from the
+  logical name instead and `Queue` reserves 5 chars (`limit=MAX_QUEUE_NAME_LENGTH -
+  len(".fifo")`); `Topic` has room to spare (80 + 8 + 5 < 256).
```
sebst, line 148: The 80 already includes the 8, no? (limit=80, default pulumi_suffix_length=8)

docs/docs/concepts/naming.md
```diff
@@ -0,0 +1,56 @@
...
+
+Changing a component's name in `stlv_app.py` is a new resource to Pulumi, not a rename.
+The next deploy creates the new one and deletes the old. The app or environment name works
+the same way, see [Renaming](state.md#renaming). Where that bites:
+
+- `Bucket`: the old bucket keeps its objects, and the delete fails while it is non-empty.
+  Move or delete the data first.
+- `UserPool`: users live in the pool. The new pool starts empty.
+- `DynamoTable`: the data goes with the table.
+- `Queue`: in-flight messages are lost with the old queue.
```
sebst, line 54: New code uses pulumi_suffix_length=0 and the test asserts a 128-char name. The new formula is right. A pool that already overflowed the old 120-char budget gets a new logical name and a new AWS name

stelvio/aws/email.py
```diff
@@ -223,7 +225,11 @@ def _create_resources(self) -> EmailResources:
                 "configuration_set",
                 {
                     "resource_name": context().prefix(f"{self.name}-config-set"),
```
sebst, line 227: should be resource_name, too?

docs/docs/changelog.md
michal-stlv: yes, 67 for fifo queues. rewrote the whole bullet in e6214f3, it was hard to read anyway.

docs/docs/contributing/components.md
michal-stlv: yes, the 80 already reserves the 8. fixed in e6214f3, 80 + 5 < 256.

docs/docs/concepts/naming.md
michal-stlv: true, and it's any pool over 120, not just 121-128: over 128 both versions truncate, but at 120 vs 128. added identity pools to the replaced list in the changelog, e6214f3. no alias here, the name input itself changes and it's force-replace, so an alias wouldn't keep the pool.

stelvio/aws/email.py
michal-stlv: it's the logical name, no aws limit. this pr only touches the sites where the aws name changes, next pr moves all remaining call sites to resource_name in one sweep.

sebst (approved):

PR #285, opened 2026-09-21
refactor: resource_name is the only naming helper

michal-stlv:
## Why

#284 added `resource_name` and used it at the nine sites whose AWS name changed. The other 54 sites still called `safe_name(context().prefix(), name, limit, ...)`, so two helpers did one job and every caller repeated the prefix. This PR finishes the move: one helper, the prefix comes from the context, `safe_name` is gone.

## What changes

Nothing for users. Every call is a 1:1 rewrite (same base, same limit, same suffix, same Pulumi reservation), so every logical and physical name is byte-identical and no deployed stack sees a change.

- `safe_name(prefix, name, max_length, suffix, pulumi_suffix_length)` deleted; its body now lives in `resource_name(base, *, limit, suffix="", pulumi_suffix_length=8)`.
- 54 call sites in 19 files rewritten. `Vpc._safe_name` is now `Vpc._resource_name`. The `prefix` locals that only fed `safe_name` are gone.
- One error text changed: "Cannot create safe name" is now "Cannot build resource name".

## Tests and docs

The helper's tests moved from `tests/test_context.py` (which held nothing else) to `tests/test_component.py`, computed against the fixture prefix. The five tests that patched `safe_name` now patch `resource_name` with keyword asserts. `contributing/components.md` §Naming loses the paragraph about the explicit-prefix form.

Unit suite green. Integration standard and CloudFront tiers green.

## Review

Two commits on purpose. The first is the mechanical sweep: after it, `grep -rn safe_name stelvio` finds only the definition. The second is the fold, the only logic change. Read that one on its own.


stelvio/component.py
sebst: worth renaming this to `resource_key` to avoid confusion?

stelvio/aws/function/function.py
sebst: see: https://github.com/stelviodev/stelvio/pull/285/changes#r4065232644

resource_name not used here, but added import adds confusion

sebst (dismissed):

stelvio/component.py
michal-stlv: Done in d845cb6: `_customizer(resource_key, ...)`.

stelvio/aws/function/function.py
michal-stlv: Same fix, d845cb6: the lambda param is `resource_key` now, so nothing shadows the imported `resource_name`.

sebst (approved):

PR #287, opened 2026-09-21
chore: update pulumi version and other dependencies

sebst:


michal-stlv (commented):
okay, great stuff, all looking good. ran unit tests, docs build and all three integration tiers against real aws on the new versions, all green. one question on the pymdown pin, see the comment.

pyproject.toml
```diff
@@ -69,6 +69,11 @@ include = [
     "/LICENSE",
 ]
 
+[tool.uv]
+constraint-dependencies = [
+    "pymdown-extensions==11.0.1",
```
michal-stlv: Why pin `pymdown-extensions` to exactly 11.0.1? `mkdocs-material` and `zensical` both accept newer versions. The lock resolves without the constraint. I tried the docs build with 12.0.1 and the output is identical. If 12 broke something for you, can we add a short comment saying what? If not, can we drop the constraint?

michal-stlv (commented):
left one comment, otherwise good, ran all integration tests, all passing

pyproject.toml
sebst: just extra cautious to not jump across two majors. dropped now

michal-stlv (approved):

PR #288, opened 2026-09-23
fix: cli diff and deploy output, failure ending, no hang after a provider error

michal-stlv:
## Why

A batch of CLI output problems seen on real deploys, all in the rich deployment handler and the CLI failure path, so one review context.

- Unchanged children under a changed parent rendered as bare dim `~ Function x` lines with nothing under them; on a partial-deploy resume five of them read as real updates. Parents of a not-yet-populated sub-component rendered as bare active lines for the first seconds of a deploy for the same reason.
- Pulumi prefixes provider warnings with the full URN; the tree already names the resource one line above, so a warning wrapped three lines.
- `--json` sent a policy document with no `Resource` key when the ARN was unknown at preview, which reads as permissions on every table.
- Top-level components rendered in arrival order, so two identical diffs could differ.
- The bucket data-loss warning never fired: the allowlist had the BucketV2 token, Stelvio creates `aws:s3/bucket:Bucket`.
- Error text containing `[...]` broke rich markup (raised, or ate the token).
- A real failure ended with a cruder `✕ Failed` from `pulumi.py`; the handler's `✗ Deployed in 12s with errors` and counts were unreachable, the counts called a failed resource "deployed", the spinner kept drawing under the error and the cursor stayed hidden.
- A delete-before-replace (explicitly named resources) emits `delete-replaced`, `replace` and `create-replacement` for one URN in no fixed order; the first pre-event wins and `delete-replaced` was mapped nowhere, so a replaced Bucket flipped between `(1 to replace)` and unchanged between two identical diffs.
- A resource failing provider Check before any step was reported as `create`, so an existing queue showed `✗ Queue tasks  (1 to create)`.
- After an engine failure `stlv` printed its ending and never exited.

## What changes

- `_is_hidden_unchanged`: SAME resources and empty sub-components are hidden under a changing parent in diff and deploy frames; `--show-unchanged` shows them; failed children always show. A component with no child operation has none (`ComponentInfo.operation -> OpType | None`), so its parent no longer reads as "create" while child events are on the way.
- Diff and refresh frames sort top-level components; deploy keeps event order so lines don't jump under the spinner.
- `_clean_diagnostic_message` cuts the event's URN; the regex fallback stops at whitespace or comma so the "interrupted while creating" detector still matches.
- `--json` update entries gain `unknown_paths` (`["Statement[0].Resource"]`), computed by the same heuristic the tree uses for `output<string>`; `old` and `new` stay byte-exact.
- `aws:s3/bucket:Bucket` joins the data-loss set and the label map.
- `escape()` on the interpolated value at the eleven `[red]…[/red]` sites and the warnings summary; tag style kept.
- `_show_simple_error` ends with `handler.show_completion(failed=True)`: `_stop_live` first, then `✗ Deployed in 4s with errors` and `1 component (2 resources) deployed, 1 failed`; no counts after a failed preview.
- `OpType.DELETE_REPLACED` is a replace everywhere (`_REPLACE_OPS`, component priority, glyph maps).
- A Check-failed resource has no operation (`ResourceInfo.operation: OpType | None`): the header drops the count, JSON and stream say `unknown`. A failed resource is always visible, including internal-managed API Gateway ones that are otherwise shown only on create.
- The hang. The Automation API's inline-program worker (a grpc server pool thread) is still running `stlv_app.py` when `stack.preview()` raises: it sits in `get_policy_document` awaiting its provider's URN, which the cancelled engine never resolves, and interpreter shutdown joins every pool worker. Pulumi's fix for pulumi/pulumi#11594 covers raised RPC errors only. New console-script entry `stelvio.cli:main` (for `stlv` and `stelvio`) runs click, catches the SystemExit click always ends with, flushes and `os._exit`s. By then every `with CommandRun` has released the state lock and removed the work dir; only atexit hooks are skipped, none of which Stelvio relies on. `_handle_error` under `STLV_DEBUG` prints the traceback instead of re-raising so it takes the same exit.

## Tests

Twenty-three new tests across the handler, CLI exit-code and pulumi test files, one to four per item, goldens updated where lines changed. `test_cli_exits_even_when_a_thread_pool_worker_never_finishes` spawns a child with a blocked pool worker and calls `main()`; the same child through the old `cli()` entry never reports an exit code. Fifteen pins flipped from `create` to `unknown`. Unit suite green.

Verified on a real stack: three identical diffs of a forced bucket replacement (`± Bucket uploads  (4 to replace)` with the data-loss line each time); a queue-timeout deploy and its restore captured on a pseudo-terminal (only the changed component in every frame); a failing deploy (`✗ Deployed in 4s with errors`, `1 component (2 resources) deployed, 1 failed`); a provider Check failure on an existing queue (`✗ Queue tasks`, exit 1 in 5s, no hang, where main hangs).


docs/docs/changelog.md
```diff
@@ -19,6 +19,7 @@
 
 - **Same-type child resources show which one they are.** `Subnet (public-a)`, `API Method (GET /users/{id}/orders)`, `IAM Policy Attachment (default)` instead of six identical `Subnet` lines.
 - **Diff output is sorted.** `stlv diff` and `stlv refresh` group children by type, sub-components first, API paths as a tree. `stlv deploy` keeps arrival order.
+- **Cleaner diff and deploy output.** No more unchanged lines under a changed component, a failed run says what failed, and the data-loss warning on replacement now covers buckets too.
```
sebst: should say that stlv used to sit there after a provider error, and that it now exits

sebst (approved):

docs/docs/changelog.md
michal-stlv: it could but i wanted to keep this short.

